PR #10, opened 2018-07-06
doc: build howto temporarily points to next in sof

mmaka1:
This additional step is necessary until next is merged back to master.


marc-hb:
Reverted in PR #199

PR #23, opened 2018-08-22
doc: remove outdated build guide

xiulipan:
Some old guide should be removed.


xiulipan:
@lgirdwood 
Remove the converting guide from WIKI.

PR #60, opened 2018-11-07
build: update listings to use smp

jajanusz:
Signed-off-by: Janusz Jankowski <janusz.jankowski@linux.intel.com>

### DIFF
--- a/developer_guides/unit_tests.rst
+++ b/developer_guides/unit_tests.rst
@@ -46,7 +46,7 @@ add **--with-cmocka-prefix=<path to cmocka>**. For example:
 .. code-block:: bash
 
    ./autogen.sh
-   ./configure --with-arch=xtensa --with-platform=apollolake \
+   ./configure --with-arch=xtensa-smp --with-platform=apollolake \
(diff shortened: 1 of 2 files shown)


lgirdwood:
@xiulipan can you add these SMP configs into CI. Thanks !

lgirdwood:
@xiulipan it's an additional build for platforms > 1 core.

jajanusz:
@lgirdwood  @xiulipan SMP should be default build for all multicore platforms (APL and above), because of having UP as default so far there are some bugs that got unnoticed by CI - like too small memory sections when building for SMP which is our default target for these platforms - https://github.com/thesofproject/sof/pull/455 broke it

I'd wait with adding this to CI until memory section fix is merged (should be today), I can change SMP to be default build in xtensa-build-all.sh when it will be merged

jajanusz:
> As this is some additional build, I think you can have more lines about this.
> Not just some simple replacement.

@xiulipan As i wrote it's not additional build, what you had before can be considered additional build for these platforms, SMP is default for APL+, UP for earlier like BYT ;)

xiulipan:
@jajanusz 
If this is some default change. I think nothing is need for CI.
You need to just change the xtensa-build-all.sh and will enable CI build check.

PR #68, opened 2018-12-06
docs: logger: add logger usage description

bkokoszx:
I've added logger usage description. It is actually ported README tools file.

Signed-off-by: Bartosz Kokoszko <bartoszx.kokoszko@linux.intel.com>


bkokoszx:
I would be thankful for every style suggestions - it is my first time using rst.

jajanusz:
looks good

developer_guides/debugability/logger/index.rst
```diff
@@ -0,0 +1,106 @@
...
+
+
+Examples:
+
+- Get traces from "/sys/kernel/debug/sof/etrace" file, verifies fw_version with
+  "/sys/kernel/debug/sof/fw_version" and prints logs to stdout 
+ 
+.. code-block:: bash
+
+	$ sof-logger -l ldc_file -e
```
jajanusz: Just one thing - indentation is 3 spaces not tab

jajanusz: Just one thing - indentation should be 3 spaces instead of tab

bkokoszx: updated

PR #74, opened 2018-12-28
build from scratch update

fredoh9:
This is required changed after soft was depreciated.

One small concern and need your feedback is
1. kvm libvirt-bin packages are not available in Ubuntu 18.04 and 18.10. I assume these are available some old version of Ubuntu. So I didn't remove them.

2. I couldn't find below path after building alsa-lib. So I deleted this line. May old alsa-lib create libaries to that path? I saw one of makefile has /usr/lib/alsa-lib path but any of alsa folder is not created.
$ sudo cp /usr/lib/alsa_lib/*    /usr/lib/x86_64-linux-gnu/alsa-lib

3. I had hard time to build alsa-utils. gitcompile didn't work for me. Not sure why that happened to me. I have good solution but I just put short comments only.

### DIFF
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -56,11 +56,11 @@ the newest ALSA from source code.
    $ ./gitcompile
    $ sudo make install
 
-Replace the default Ubuntu alsa-lib with the one we just built.
+Update an environment variable to refer to the alsa-lib with the one we just built.
 
 .. code-block:: bash
 
-   $ sudo cp /usr/lib/libasound.*    /usr/lib/x86_64-linux-gnu/
+   $ export LD_LIBRARY_PATH=~/work/sof/alsa-lib/src/.libs:$LD_LIBRARY_PATH
 
 Clone, build, and install alsa-utils.
 
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -28,7 +28,9 @@ Install package dependencies.
 .. code-block:: bash
 
    $ sudo apt-get install libgtk-3-dev libsdl-dev libspice-protocol-dev libspice-server-dev libusb-1.0-0-dev libusbredirhost-dev \
-                        libtool-bin iasl valgrind texinfo virt-manager kvm libvirt-bin virtinst libfdt-dev libssl-dev pkg-config
+                        libtool-bin iasl valgrind texinfo virt-manager kvm libvirt-bin virtinst libfdt-dev libssl-dev pkg-config \
+                        help2man gawk libncurses5 libncurses5-dev
+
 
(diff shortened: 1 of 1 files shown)


getting_started/build-guide/build-from-scratch.rst
```diff
@@ -155,43 +160,45 @@ Build and install the headers for each platform.
...
    $ make
    $ make install
    #Haswell
-   $ ./configure --target=xtensa-hsw-elf --prefix=~/work/sof/xtensa-root
+   $ ./configure --target=xtensa-hsw-elf --prefix=/home/$USER/work/sof/xtensa-root
    $ make
    $ make install
    #Apollo Lake
-   $ ./configure --target=xtensa-apl-elf --prefix=~/work/sof/xtensa-root
+   $ ./configure --target=xtensa-apl-elf --prefix=/home/$USER/work/sof/xtensa-root
```
lgirdwood: What's the reason for changing ~/ here ?

fredoh9: configure script expects full path not relative path. Maybe that requirement was changed after the document.

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -59,7 +61,6 @@ Replace the default Ubuntu alsa-lib with the one we just built.
 .. code-block:: bash
 
    $ sudo cp /usr/lib/libasound.*    /usr/lib/x86_64-linux-gnu/
-   $ sudo cp /usr/lib/alsa_lib/*    /usr/lib/x86_64-linux-gnu/alsa-lib
```
plbossart: this will essentially prevent pulseaudio from working on the host machine. It's better to use LD_LIBRARY_PATH=<path-to-alsa-lib>/src/.libs:$LD_LIBRARY_PATH.

fredoh9: HI Pierre,

Changing the line is not big deal. Trying to clarify your comments. You meant removing both lines and adding exporting LD_LIBRARY is better way even considering PulseAudio? Bear with me, I have little idea of PulaseAudio. I appreciate your comments but not sure why this will break PulseAudio. Could you explain a bit more?

plbossart: The point was to avoid breaking the host setup with a wild install of alsa-lib, so yes the two lines should be removed.
There are two ways to make sure the latest alsa-lib is used:
1) don't install but only use an environment variable to refer to the latest alsa-lib
2) build a debian/rpm package and install it cleanly.

fredoh9: Now 100% clear to me. new commit is made.

https://github.com/thesofproject/sof-docs/pull/74/commits/93948b8af7c5c13c79351013a87c89a892eed4ba

lgirdwood:
All comments addressed, Thanks.

Issue #86, opened 2019-01-30
remove the prompt in all bash examples


plbossart:
Make sure people can just copy/paste instructions in a terminal window for all bash examples

Instructions such as
````bash
$ sudo add-apt-repository ppa:ubuntu-toolchain-r/test
$ sudo apt-get update
$ sudo apt-get install gcc-7 g++-7
$ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 70 --slave /usr/bin/g++ g++ /usr/bin/g++-7
````
should really be 

````bash
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-7 g++-7
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 70 --slave /usr/bin/g++ g++ /usr/bin/g++-7
````

**Comment by intelkevinputnam** [comment]
Good point. Totally agree.

**Comment by stripes416** [comment]
Could I take this issue? I don't seem to be able to assign it to myself at the moment.

**Comment by intelkevinputnam** [comment]
@stripes416 You've got it. :)

**Comment by stripes416** [comment]
Thanks @intelkevinputnam!

**Comment by plbossart** [comment]
> Could I take this issue? I don't seem to be able to assign it to 
> myself at the moment.
>
In addition to the fix of the documentation, we should have a matching 
script that does everything for you. Instructions are always broken for 
months before people bother to report them, a broken script will trigger 
more reports.

The hard-coded paths should go away, e.g. ~/work/sof should be replaced 
by a variable that can be easily adjusted.

Thanks

-Pierre

**Comment by lgirdwood** [comment]
@stripes416 whilst you are updating docs the following steps also need rectified

1) Building xtensa-newlib needs a
```
rm -fr rm etc/config.cache
```
after make install and before configure of next target.

2)  The install deps are different for Ubuntu 18.10 compared to Ubuntu 18.4 (documented). + flex bison - kvm e.g.
```
udo apt-get install libgtk-3-dev libsdl-dev libspice-protocol-dev libspice-server-dev libusb-1.0-0-dev libusbredirhost-dev                      libtool-bin iasl valgrind texinfo virt-manager virtinst libfdt-dev libssl-dev pkg-config                      help2man gawk libncurses5 libncurses5-dev flex bison
```

stripes416:
[...]

PR #92, opened 2019-02-06
Remove prompt code examples

stripes416:
Addresses #86 -- prompt removal from code examples.
Other requests are addressed in PR #93

### DIFF
--- a/getting_started/setup/setup_up_2_board.rst
+++ b/getting_started/setup/setup_up_2_board.rst
@@ -66,9 +66,9 @@ Copy kconfig fragments for SOF development from https://github.com/thesofproject
 
 .. code-block:: bash
 
-   $ make defconfig
-   $ scripts/kconfig/merge_config.sh .config <path>/kconfig/base-defconfig <path>/kconfig/sof-defconfig  <path>/kconfig/hdaudio-codecs-defconfig
-   $ make -j8
+   make defconfig
+   scripts/kconfig/merge_config.sh .config <path>/kconfig/base-defconfig <path>/kconfig/sof-defconfig  <path>/kconfig/hdaudio-codecs-defconfig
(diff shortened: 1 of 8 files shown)


stripes416:
I don't really know what the standard for the project is as far as multi-commit vs squashed PRs, but please let me know if I need to make those kind of changes.

getting_started/setup/setup_up_2_board.rst
```diff
@@ -66,9 +66,9 @@ Copy kconfig fragments for SOF development from https://github.com/thesofproject
 
 .. code-block:: bash
 
-   $ make defconfig
-   $ scripts/kconfig/merge_config.sh .config <path>/kconfig/base-defconfig <path>/kconfig/sof-defconfig  <path>/kconfig/hdaudio-codecs-defconfig
-   $ make -j8
+   make defconfig
+   scripts/kconfig/merge_config.sh .config <path>/kconfig/base-defconfig <path>/kconfig/sof-defconfig  <path>/kconfig/hdaudio-codecs-defconfig
```
intelkevinputnam: Need to be careful about those angle brackets. Copy and pasting them wouldn't work out very well. OK for now, I guess, though in the long term I think a better solution is to tell the user to set a variable and use the variable in the instructions.

xiulipan:
@stripes416 
Thanks for helping!

getting_started/setup/setup_up_2_board.rst
stripes416: @intelkevinputnam I agree --- I think you referenced this solution in #86. I'll work on it and open a PR.

intelkevinputnam: No worries. I just like to beat dead horses.

PR #95, opened 2019-02-21
cmake: add openssl path argument

jajanusz:
Updated for new argument + Added some sections descriptions

Signed-off-by: Janusz Jankowski <janusz.jankowski@linux.intel.com>

### DIFF
--- a/developer_guides/cmake.rst
+++ b/developer_guides/cmake.rst
@@ -3,8 +3,15 @@
 CMake Arguments
 ###############
 
+For firmware and unit tests only **TOOLCHAIN** and **ROOT_DIR**
+arguments are mandatory. Other arguments are optional.
+
+For host build, only **BUILD_HOST** switch is needed.
+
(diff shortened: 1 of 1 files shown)


developer_guides/cmake.rst
```diff
@@ -54,9 +61,21 @@ MEU_PRIVATE_KEY
       # Example
       cmake [...] -DMEU_PRIVATE_KEY=/path/to/meu/private-key.pem [...]
 
+MEU_OPENSSL
```
lgirdwood: Can we mark this as optional and as Intel and Windows only since we now are merging PRs for non Intel architectures.

jajanusz: All variables are optional. Only `TOOLCHAIN` and `ROOT_DIR` are needed if  you build FW

jajanusz:
@lgirdwood 
Updated with more info about which arguments are mandatory/optional

PR #121, opened 2019-05-10
Minor doc updates

plbossart:
updated links for kconfigs and added info on supported platforms (included legacy)

### DIFF
--- a/platforms/intel-cavs/index.rst
+++ b/platforms/intel-cavs/index.rst
@@ -5,13 +5,17 @@ Intel CAVS Platforms
 
 Intel CAVS platforms supported by the |SOF|.
 
-+---------+----------+--------------------+
-|cAVS     |ver. 1.5  | Apollo Lake        |
-|         +----------+--------------------+
-|         |ver. 1.8  | Cannon Lake        |
-|         +----------+--------------------+
-|         |ver. 2.0  | Ice Lake           |
-+---------+----------+--------------------+
++---------+----------+-----------------------------------------+
+|cAVS     |ver. 1.5  | Apollo Lake, Gemini Lake                |
+|         +----------+-----------------------------------------+
+|         |ver. 1.8  | Cannon Lake, Whisky Lake, Comet Lake    |
+|         +----------+-----------------------------------------+
+|         |ver. 2.0  | Ice Lake                                |
++---------+----------+-----------------------------------------+
+
+.. note:: while the Sky Lake and Kaby Lake platforms are also based on the
+	  cAVS 1.5 architecture, they are not supported at this time due to
+	  differences in boot flow and memory architecture.
 
 .. uml:: images/cavs-platform-deps.pu
    :caption: CAVS Platforms
--- a/platforms/intel-legacy/broadwell/index.rst
+++ b/platforms/intel-legacy/broadwell/index.rst
@@ -1,0 +1,13 @@
+.. _platform-broadwell:
+
+Intel Broadwell
+###############
+
+Intel Broadwell platform is based on an core CPU and a Tensilica HiFi2
+EP DSP integrated in the PCH. It relies only on internal memory (no
+caches).
+
+Supported commercially-available platforms are limited. The SOF team
+is only aware of the Chromebook Pixel 2015 (code name "samus") and the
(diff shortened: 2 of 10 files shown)


platforms/intel-legacy/broadwell/index.rst
```diff
@@ -0,0 +1,13 @@
+.. _platform-broadwell:
+
+Intel Broadwell
+###############
+
+Intel Broadwell platform is based on an core CPU and a Tensilica HiFi
+EP DSP integrated in the PCH. It relies on both internal memory (no
```
intelkevinputnam: Line 7: Not sure what is meant here: "... relies on both internal ... "

intelkevinputnam: Likely meant "only" instead of "on both".

platforms/intel-cavs/index.rst
```diff
@@ -5,13 +5,17 @@ Intel CAVS Platforms
...
-+---------+----------+--------------------+
++---------+----------+-----------------------------------------+
+|cAVS     |ver. 1.5  | Apollo Lake, Gemini Lake                |
+|         +----------+-----------------------------------------+
+|         |ver. 1.8  | Cannon Lake, Whisky Lake, Comet Lake    |
+|         +----------+-----------------------------------------+
+|         |ver. 2.0  | Ice Lake                                |
++---------+----------+-----------------------------------------+
+
+Note: while the Sky Lake and Kaby Lake platforms are also based on the
```
intelkevinputnam: Line 16: Should use the note directive: .. note::

platforms/intel-legacy/index.rst
```diff
@@ -0,0 +1,21 @@
+.. _platforms-intel-Hifi-EP:
+
+Intel Hifi-EP Platforms
```
lgirdwood: iirc, this is HiFi2 EP

platforms/intel-legacy/broadwell/index.rst
plbossart: yes, copy/paste, it's 'only'

platforms/intel-legacy/index.rst
plbossart: yes

plbossart:
> Is it worth adding a small section for "Haswell, Skylake and Kabylake". This would just state that FW can be built for these platforms but it's unverified and not officially supported. We could add for HSW that it's due to lack of any known Haswell laptop with DSP mode enabled and SKL/KBL code signing (alongside the boot differences flow on driver side) requirements make it impractical atm.
> Not forgetting Sue Creek, a single sentence here could describe it as WIP, that's currently unsupported but verified booting.

well, I don't see the point of describing a Haswell platform that doesn't seem to exist commercially and isn't supported.
Likewise for SueCreek, we took the code out and there's currently no active developer, so it's misleading to state a WIP state.

lgirdwood:
@plbossart fair enough for haswell, but sue creek does exist and could be completed by community. i.e we need a table like the https://nouveau.freedesktop.org/wiki/FeatureMatrix/

This would show each platform and it's feature status, making it obvious that sue creek will build, should boot, should have working DW DMA, should have working SSP but has "stalled" and is a "community" project. Likewise for Hikey960. That would allowed interested parties to take over, work on it and upstream where the matrix could be updated.

plbossart:
On 5/24/19 5:03 AM, Liam Girdwood wrote:
> @plbossart <https://github.com/plbossart> fair enough for haswell, but 
> sue creek does exist and could be completed by community. i.e we need a 
> table like the https://nouveau.freedesktop.org/wiki/FeatureMatrix/

There is no driver for Sue Creek

> This would show each platform and it's feature status, making it obvious 
> that sue creek will build, should boot, should have working DW DMA, 
> should have working SSP but has "stalled" and is a "community" project. 
> Likewise for Hikey960. That would allowed interested parties to take 
> over, work on it and upstream where the matrix could be updated.

I don't mind showing a WIP status but that is a different topic/section.
Supported platforms means people can start using them - possibly file 
bugs against non-working capabilities - but the bulk of the work is done 
and there are people handling such bugs.

PR #124, opened 2019-05-23
api: link to component api dox added

mmaka1:

### DIFF
--- a/api/component-api.rst
+++ b/api/component-api.rst
@@ -1,0 +1,9 @@
+.. _component-api:
+
+Component API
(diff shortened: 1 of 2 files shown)


dbaluta:
@mmaka1 can you elaborate a little bit more about why we need this commit. For an outside reader it is not clear what is the value brought with this change.

Also, please use the imperative mood in the subject line.

E.g api: Add link to doxygen component API ?

mmaka1:
@dbaluta. Fixed.

PR #138, opened 2019-10-30
platforms: Update info for Broadwell

dpward:
Because of kernel bug 204237 (and downstream Fedora bug 1750194), we know about more devices which use this platform.

### DIFF
--- a/platforms/intel-legacy/broadwell/index.rst
+++ b/platforms/intel-legacy/broadwell/index.rst
@@ -7,7 +7,11 @@ Intel Broadwell platform is based on an core CPU and a Tensilica HiFi2
 EP DSP integrated in the PCH. It relies only on internal memory (no
 caches).
 
-Supported commercially-available platforms are limited. The SOF team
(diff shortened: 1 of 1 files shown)


platforms/intel-legacy/broadwell/index.rst
```diff
@@ -8,6 +8,6 @@ EP DSP integrated in the PCH. It relies only on internal memory (no
 caches).
 
 Supported commercially-available platforms are limited. The SOF team
-is only aware of the Chromebook Pixel 2015 (code name "samus") and the
-Dell XPS.
+is aware of the Chromebook Pixel 2015 (code name "samus") and a few
+tablets and convertibles from Dell and HP.
```
plbossart: such as


    Dell Latitude 7350 - convertible
    Dell Venue 11 Pro 7140 - tablet
    HP Spectre x360 (2015) - convertible

?

dpward: Sure, I changed it to list all known devices (and I used the manufacturer's name for each).

PR #154, opened 2019-12-31
alsa-lib build linking errors

mwasko:


getting_started/build-guide/build-from-scratch.rst
```diff
@@ -101,6 +101,17 @@ Clone, build, and install alsa-utils.
...
+
+.. code-block:: bash
+
+   cd ../alsa-lib
+   ./gitcompile --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu/
+   sudo make install
+   cd ../alsa-utils
+   ./gitcompile --prefix=/usr --with-curses=ncurses --disable-xmlto --disable-bat
+   sudo make install
+
```
plbossart: we shouldn't recommend a make install in the first place, it's a terrible idea to override the distro settings. we should use /usr/local here.

lgirdwood: @mwasko @plbossart fwiw we can easily use custom alsa versions (alongside overiding other common tooling) in our docker container. This shoudl be the preffered way too, as it doesn't break a developers host PC.

mwasko: @plbossart current build from scratch guide recommend to make install and my tips only provide info how to resolve eventual ALSA compile linking errors. I can add some warning message that this operation will override system ALSA library.

There is always an option to use docker as @lgirdwood  wrote.

lgirdwood: warning message is needed. Lets the users know this will overwrite their ALSA install and will likely break things and should only be done if they know what they are doing etc. Preference is to install ALSA locally or use docker.

deb-intel: @lgirdwood Is everyone in agreement with the changes?

mwasko: @deb-intel as recommended I have added commit with warning note related to alsa-lib install operation

deb-intel: Thanks @mwasko. I will merge this when it is approved by @plbossart or @lgirdwood.

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -84,6 +84,10 @@ the newest ALSA from source code.
    ./gitcompile
    sudo make install
 
+.. note::
+
+   | The alsa-lib install operation will override your distro alsa libraries! 
+   | If you don't want to do that then please check "Build SOF with Docker" chapter.
```
lgirdwood: Probably best to put the warning before the ```make install``` rule otherwise some folks may just copy and paste blindly as they read down and nuke there alsa install. I would probably word it as follows too.
```
**WARNING** installing alsa-lib system wide may break some audio applications. 
Please only do this if you know what you are doing. It's recommended to install it locally (under $HOME)
or use Docker (see "Build SOF with Docker" chapter.)

mwasko: @lgirdwood commit with warning message v2 added

lgirdwood:
@mwasko thanks, btw should these two patches be squashed ?

mwasko:
@lgirdwood I was planning to keep these two patches separate.

deb-intel:
@plbossart Do you approve these changes?

PR #155, opened 2020-01-08
Add a SOF VirtIO design document

lyakh:
This document describes design of the SOF VirtIO / vhost
virtualisation solution.

Unfortunately, I haven't been able to build it, the error message wasn't clear.


developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
+.. _virtualisation:
+
+SOF VirtIO design
+#################
+
+.. contents::
+
+1. General information
+**********************
+
```
lgirdwood: I would add something here describing the feature. i.e. what does this code do and why it is useful.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+VirtIO is a platform-agnostic virtualisation standard, managed by OASIS
+https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=virtio . It defines
+a design and a protocol for implementing device IO in VMs.
+
+When Linux is running in a VM, multiple drivers are already available, using
+virtio: networking, storage, etc. Those drivers use the virtio and virtqueue
+APIs to communicate with the host or, potentially, with other guests.
+
+The host part depends on the used virtualisation solution. Some of them
```
lgirdwood: You should describe the "host" and "guest" concepts prior to this paragraph, I would do this in the "what the code does" section right at the start.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+subdirectories in the Linux kernel, e.g. drivers/net, drivers/block etc. On the
+contrary, all vhost drivers are located under drivers/vhost. That directory also
+contains vhost.c and vhost.h which are used by all vhost drivers.
+
+The SOF virtualisation solution also implements a vhost driver.
+
+2. SOF VirtIO design
+********************
+
+The principal point of the SOF VirtIO design is using the IPC API, that is also
```
lgirdwood: principal point is to virtualize the IPC command and audio data flows between host and guests. I would introduce the doorbell, mailbox and zero copy (for audio data and stream position) concepts here too

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+---------------------------
+
+The goal of the SOF virtualisation is to implement low latency audio
+virtualisation, which means avoiding copying of audio data and stream position
+updates. In this case only one VirtQueue would be required for IPC message
+exchange. However, during the first stage we also transfer data and stream
+position updates via VirtQueues. That means, that currently the SOF uses 3
+VirtQueues: for IPC, audio data, and stream position updates.
+
+TODO: even though currently both playback and capture are supported, they aren't
```
lgirdwood: This information is status, it's not needed in design documentation.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+ASoC / DAPM subsystems on the host are unaware of any guests, but they should be
+aware of any streaming, taking place on host parts of VM pipelines for
+configuration and power management. That means, e.g. if a guest initialises
+playback, the ASoC subsystem on the host should be aware, that the pipeline,
+connecting the respective mixer and the used codec is currently performing audio
+playback. To achieve this, for each such guest connection point we add a virtual
+PCM as a widget of type "aif_in" or "aif_out" for playback and capture
+respectively, and a virtual 0-size buffer. They are only visible on the host and
+are not sent to the DSP.
+
```
lgirdwood: Some diagram here would be useful, you can get some from my slides.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+When running in a VM, the SOF also uses ALSA topology, but in this case the
+topology should have no relation to the host hardware. The guest audio topology
```
lgirdwood: guest topology has no knowledge of HW, but host does. HW is privileged so guest access is controlled via host.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+the firmware.
+
+3. Implementation
+*****************
+
+3.1. Initialisation
+-------------------
+
+The SOF driver probing on the host remains unmodified, except additionally the
+vhost driver is initialised, which registers a /dev/vhost-dsp misc device.
```
lgirdwood: why, please explain why this is registered and what it's used for.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+TODO: define a reliable secure topology selection protocol. Currently the guest
+simply sends a configured ID to the host, upon which the host opens the topology
+file with the same name. Currently this isn't a problem, since so far only a
+single guest is supported. So, the host can verify, that the ID, that the guest
+has sent does indeed match with the guest topology file name. However, when
+support for multiple guests is added, it must be ascertained, that guests only
+receive topology files, that they are allowed to get. At the moment there
+doesn't seem to be a way on the host to uniquely and securely identify guests.
+If there is indeed no such way, our only possibility seems to be using
+cryptography for that.
```
lgirdwood: My expectation is that host FS will contain topology files for guest0, guest1, etc. The guest should identify itself somehow to the host ?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+This value is then assigned to the .ref_comp_id field of struct
+sof_ipc_comp_config during topology file parsing and sent by the guest to the
+host. The host then uses that value to modify connection IPC messages from the
+guest before sending them to the DSP.
+
+3.2. Streaming
+--------------
+
+Guest audio streaming is mostly transparent for the host. The host audio
+subsystem doesn't get involved with most guest streaming or kcontrol operations.
```
lgirdwood: please explain why.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+3.3. Deactivation
+-----------------
+
+When the guest isn't actively using audio, we want to allow the host to
+runtime suspend the DSP. That usually means switching the DSP off. Next time the
+guest wants to activate audio it will have to perform initialisation again,
+including sending its topology to the host. This happens automatically as long
+as the guest runtime suspends too or performs the initialisation cycle in some
+other way, by unloading and reloading drivers, or by rebooting.
+
```
lgirdwood: guest should cache it's topology so it does not need to ask for it from host at PM resume.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+With automatic runtime suspending on the host and the guest, it can happen,
+that one of them got suspended, whereas the other one is still running. A
+problematic case would be, if the host got suspended, while the guest was still
+active, because then, when the host resumes, it will also bring up the DSP and
+only send the host topology part to it. The guest however wasn't suspended,
+therefore it won't send its topology to the host. To avoid this the automatic
+runtime suspend timeout on the guest is set to a half of the host's timeout.
```
lgirdwood: any active guest streams should increase host PM runtime usage count.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+then forwards that topology to the DSP as usual, the DSP firmware will detect
+component duplication and enter an erroneous state. These cases also cannot be
+reliably avoided, because it's always possible, that while the guest's audio is
+idle, the host's audio is still active, e.g. being in use locally. To handle
+such cases correctly, we implement a reset count. The host counts DSP resets,
+and when the guest is resuming, it requests a reset status, using a newly added
+SOF_IPC_PM_VFE_RESET_STATUS IPC message. If the guest identifies, that the host
+hasn't been reset since the guest's last suspend, it skips sending the parsed
+topology to it.
+
```
lgirdwood: My expectation is that the guest would also send a PM IPC to the DSP to indicate its about to enter D3, the host could use this to track state ?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+the moment there is no standard way to get notified about the guest reboot event
+on the host. Various vhost drivers implement this in their own context-specific
+ways. E.g. the vhost networking driver uses a link status update for that
+purpose. However the vsock vhost driver comes very close to our needs. It
+doesn't have a suitable context notification, therefore it implements a
+dedicated QEMU / misc-device ioctl VHOST_VSOCK_SET_RUNNING. We reuse the ioctl
+definition, but change its name to VHOST_SET_RUNNING. This way no existing apps
+have to be modified. Then we use this ioctl() in QEMU SOF code to inform the SOF
+vhost driver about a guest reboot. This is then used to reset VirtQueue status
+on the host.
```
lgirdwood: We also need to teardown and guest topology on the DSP.

developer_guides/virtualisation/virtualisation.rst
lyakh: That's exactly the point - I don't think there's a standard way to securely do that. We are trying to protect against "fraud," right? I.e. someone installing a rogue VM, which would claim to be "guest0," where in fact it is not. So, as far as I understand the only way to do that is using cryptography. E.g. the host would send a request "identify yourself" along with a pseudo random number as a challenge, the guest could return a signed response "I am guest 0, your challenge was X." But I'm not a cryptography expert and I don't know how much security we really need for this.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood It does cache it of course. It resends it to the host - the host doesn't cache topology fragments of all guests, don't think that would make sense. And, of course, when the guest reboots, or reloads the driver, it requests the topology again.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood it does, sure. But after the guest stopped streaming, it can be automatically suspended, as well as the host.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood yes, I guess, the host could use that too. But the point is, that the guest has to know, whether the DSP was suspended since it last was using it. Otherwise it will send the topology to the host when it resumes, even if the DSP wasn't suspended, and therefore still has the complete topology.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood is there a way to do that apart from resetting the DSP? Otherwise - is there any harm in it keeping the topology until the next reset, whenever that happens?

developer_guides/virtualisation/virtualisation.rst
lgirdwood: Lets leave this open atm, guest ID will be trusted so no need for any crypto challenge. Leave room in IPC for guest ID in any init flow.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: yep, there is IPC to teardown pipelines. Your pipeline will be in guest component IPC ID range

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+SOF VirtIO design
+#################
+
+.. contents::
+
+1. General information
+**********************
+
+VirtIO is a platform-agnostic virtualisation standard, managed by OASIS
+https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=virtio . It defines
```
plbossart: tag for http for highlighted links?

developer_guides/virtualisation/virtualisation.rst
plbossart: +1

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+a design and a protocol for implementing device IO in VMs.
+
+When Linux is running in a VM, multiple drivers are already available, using
+virtio: networking, storage, etc. Those drivers use the virtio and virtqueue
+APIs to communicate with the host or, potentially, with other guests.
+
+The host part depends on the used virtualisation solution. Some of them
+implement the VirtIO API themselves, then using their method of choice to talk
+to the whatever host they are running on. A typical Linux virtualisation
```
plbossart: to the host

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+implementation uses QEMU + KVM to run guest instances. In this case there are
+two ways to implement a VirtIO host: in the user space and in the kernel. If
+implemented in the user space, it is managed by QEMU. In the kernel space VirtIO
+host drivers use the Linux vhost API.
```
plbossart: add a diagram and clarify if we support both modes?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+It is also worth to note, that VirtIO guest drivers reside in respective driver
+subdirectories in the Linux kernel, e.g. drivers/net, drivers/block etc. On the
+contrary, all vhost drivers are located under drivers/vhost. That directory also
+contains vhost.c and vhost.h which are used by all vhost drivers.
+
+The SOF virtualisation solution also implements a vhost driver.
```
plbossart: do you mean to say the virtualization solution implements both a guest and vhost driver?

developer_guides/virtualisation/virtualisation.rst
plbossart: which IPC? the SOF-defined IPC?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+2. SOF VirtIO design
+********************
+
+The principal point of the SOF VirtIO design is using the IPC API, that is also
+used by the native SOF driver to communicate with the SOF code, running on the
```
plbossart: communicate with the SOF firmware (code is too vague)

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+audio DSP. We use the same IPC structures and we add 3 more VirtIO-specific IPC
+messages, which are never sent to the DSP, so the firmware doesn't need to
+support them. For the most part IPCs from guests are forwarded one-to-one by the
+host driver to the DSP.
```
plbossart: you are saying too much or too little with your last sentence. In which cases are the IPCs from guests not forwarded to the DSP.

And did you mean forwarded without any change, as is, or relayed with possibly some re-formatting/changes in some fields.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+2.1. Communication channels
+---------------------------
+
+The goal of the SOF virtualisation is to implement low latency audio
+virtualisation, which means avoiding copying of audio data and stream position
+updates. In this case only one VirtQueue would be required for IPC message
```
plbossart: low latency also means avoid context changes as much as possible.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+exchange. However, during the first stage we also transfer data and stream
```
plbossart: first stage? Did you mean initial implementation or first stage of the IPC?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+position updates via VirtQueues. That means, that currently the SOF uses 3
+VirtQueues: for IPC, audio data, and stream position updates.
+
+TODO: even though currently both playback and capture are supported, they aren't
+supported simultaneously. This should be fixed in future versions.
+
+2.2. Topology
+-------------
+
+When running natively, the SOF uses the ALSA topology API to configure audio
```
plbossart: the SOF driver uses

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+pipelines on the DSP. The SOF driver reads a platform-specific topology file
+from the filesystem, parses it and sends all the components from it to the DSP
```
plbossart: sends from it to the DSP firmware -> you really mean instantiate, configure and enable components on the DSP side.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+firmware.
+
+When configured to support virtualised audio, the host topology additionally
+provides connection points for guest audio pipelines. Guest playback pipelines
+connect to mixer and capture pipelines to demultiplexer components. The ALSA /
```
plbossart: not necessarily. You may want to dedicate a DAI to a guest. or run completely different topologies. It's not correct to assume the sound from a guest would always be mixed.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+ASoC / DAPM subsystems on the host are unaware of any guests, but they should be
+aware of any streaming, taking place on host parts of VM pipelines for
+configuration and power management. That means, e.g. if a guest initialises
+playback, the ASoC subsystem on the host should be aware, that the pipeline,
+connecting the respective mixer and the used codec is currently performing audio
+playback. To achieve this, for each such guest connection point we add a virtual
+PCM as a widget of type "aif_in" or "aif_out" for playback and capture
+respectively, and a virtual 0-size buffer. They are only visible on the host and
```
plbossart: why do you need a virtual 0-size buffer? Why isn't an input and output enough?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+are not sent to the DSP.
+
+When running in a VM, the SOF also uses ALSA topology, but in this case the
```
plbossart: the SOF what? guest driver?

developer_guides/virtualisation/virtualisation.rst
plbossart: but to be clear you can have a guest who's the sole user of a DAI, under control of the host. There should be no assumption that we go through a mixer/demux

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+topology should have no relation to the host hardware. The guest audio topology
+should be decided upon by the host system administrator. Therefore we let the
+SOF instance, running on the guest, obtain topology from the host. Such a
+topology should only contain "software components," like PCMs, buffers, software
+volume controls. Those components thus represent partial pipelines, that then
+will be linked to mixer and demux components of the host SOF topology for
+playback and capture respectively.
```
plbossart: no, please don't assume the presence of mixers. For example you could have one guest taking care of BT audio. You would never have any sound coming from the host on that link.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+the host topology. The vhost driver manages connection points between the host
+and the guests' topology fragments.
+
+guest: only "sees" the guest parts.
+
+As mentioned above, SOF IPC messages from guests are forwarded exactly 1-to-1 to
+the DSP. E.g., if a guest sends an IPC to adjust audio volume, it will be
+forwarded to the DSP. That IPC contains the volume component ID as seen by the
+respective guest. Therefore it should be exactly the same ID, as the one used by
+the firmware.
```
plbossart: but to be clear the guest can only adjust the volume in parts it sees, it cannot change the master volume and blow the speakers.

developer_guides/virtualisation/virtualisation.rst
plbossart: yes, I have no idea what a vhost-dsp misc device is, and if this is a new component you created or something already supported by virtio

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+*****************
+
+3.1. Initialisation
+-------------------
+
+The SOF driver probing on the host remains unmodified, except additionally the
+vhost driver is initialised, which registers a /dev/vhost-dsp misc device.
+
+When QEMU starts a guest instance, it detects the above misc device and
+creates a PCI device for the guest with dedicated vendor and device IDs.
```
plbossart: so now I am confused. earlier you mentioned two modes with QEMU in userspace and without QEMU, so which mode are we supporting?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+The SOF VirtIO driver on the guest registers support for that PCI device and
+its .probe() method is called.
```
plbossart: what PCI ID is this referring to? Is this the same as the cAVS IP (00:1f.3:) or a new one specific for virtualization?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+The SOF driver on the guest performs partial initialisation, it skips any
+steps, involving communicating with the actual DSP, e.g. sending the firmware,
+booting it, tracing initialisation.
```
plbossart: you should clarify that the guest cannot enable specific post-processing libraries on its own, all the libraries need to be handled by the host.

developer_guides/virtualisation/virtualisation.rst
plbossart: well, it's a major open. If there is no way to identify a guest, then this solution needs to be restricted to a single guest.
Can't we check with virtualization experts what is needed here?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+support for multiple guests is added, it must be ascertained, that guests only
+receive topology files, that they are allowed to get. At the moment there
+doesn't seem to be a way on the host to uniquely and securely identify guests.
+If there is indeed no such way, our only possibility seems to be using
+cryptography for that.
+
+Guest topology files contain virtual DAI components of type "dai_out" and
+"dai_in." Their data tuples contain a new token of type SOF_TKN_COMP_REF, which
+contains an ID of the mixer or the demux, to which this pipeline should be
+connected.
```
plbossart: no, no mixer assumption please!

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+After obtaining the topology, the guest SOF instance uses a second new
+SOF_IPC_TPLG_VFE_COMP_ID VirtIO-specific IPC message to request its allocated
+component ID base. The host maintains a bitmap of guest component ID ranges to
+guarantee, that component IDs stay unique regardless of the order in which
+guests are brought up and down.
```
plbossart: I am confused by the mix of bitmap and ranges. What are you trying to say?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Next the guest proceeds by sending the parsed topology over IPC to the host,
+which is then forwarded to the DSP firmware. The guest uses the .send_msg()
+callback to forward any IPC, including topology related, with no processing at
+all, to the host. The only exception is blocking IPCs when the guest is
+resuming, while the host doesn't need to be resumed (see 3.3 section on "reset
```
plbossart: tags for section, LaTeX style ;-)

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+count" for details).
+
+The host receives the guest topology over the IPC virtual queue. In most cases
+the host IPC handler just forwards IPCs to the DSP, but in some cases it has to
+handle or modify IPC contents locally. One such example is linking the guest and
+host topology fragments together. As mentioned above, guest topology files
+contain a new token, referencing mixer or demux components in the host topology.
```
plbossart: avoid dependency on mixer/demux

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+This value is then assigned to the .ref_comp_id field of struct
+sof_ipc_comp_config during topology file parsing and sent by the guest to the
+host. The host then uses that value to modify connection IPC messages from the
+guest before sending them to the DSP.
```
plbossart: do you mean to say that there is a guest-ID to host-ID translation?

developer_guides/virtualisation/virtualisation.rst
plbossart: shouldn't the host allocate the DMA buffers and forward interrupts?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+3.2. Streaming
+--------------
+
+Guest audio streaming is mostly transparent for the host. The host audio
+subsystem doesn't get involved with most guest streaming or kcontrol operations.
+All those operations are completely processed on the guest and only resulting
+DSP IPCs are forwarded to the host. The only exceptions are beginning and ending
+of the streaming, at which times the ASoC subsystem on the host has to be
+informed, that the respective virtual pipeline, typically including a physical
+codec DAI, is becoming active or inactive respectively. Deactivation of the
```
plbossart: this is super confusing. Earlier you mentioned that the guest only sees parts of the topology, so I don't get the concept of 'virtualized pipeline' that includes hardware DAIs. 

Please clarify why you mean by 'virtualized pipeline', 'guest pipeline'.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+guest pipeline on the host requires changes to the ASoC core, without which the
+guest audio path is not recognised as inactive and is not pruned, thus making
+further streaming by the guest impossible.
+
+Upon reception of the STREAM_TRIG_START IPC message the vhost driver activates
+the virtual PCM pipeline and updates the routing information on the host.
```
plbossart: what 'routing information' are you referring to? Also the vhost driver is already part of the host, so what update are you referring to?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+As mentioned above, it is our goal to implement data zero-copy. As long as
+this isn't the case, the guest VirtIO driver implements .block_write() and
```
plbossart: You should really have two models
- copy over virtio queues
- zero-copy

and a means to switch between the two for debug.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Similarly to the data, streaming buffer position updates are currently also
+transferred via a dedicated VirtQueue. Both the host and the guest are
+configured to use IPC messages for position updates. Buffers for VirtIO
+VirtQueues are always provided by guests, therefore there should always be a
+position update buffer waiting on the host side to send a position update
+message as soon as one arrives from the DSP.
+
+Upon reception of the STREAM_TRIG_STOP IPC message the vhost driver updates
+the routing information and deactivates the virtual PCM pipeline on the host.
```
plbossart: again this is far from clear, please clarify why routing information and virtual pipeline refer to.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,255 @@
...
+
+	soc_dpcm_runtime_update();
+	snd_soc_runtime_deactivate();
+
+doesn't deactivate the pipeline and therefore doesn't allow runtime PM to
+suspend the interface, which then breaks following activation attempts. The
+current VirtIO implementation contains a fix for that, but that fix might turn
+out to be incorrect. The difficulty in establishing a correct fix is, that so
+far it hasn't been possible to find a test case for
+snd_soc_runtime_deactivate().
```
plbossart: can't follow you here.

developer_guides/virtualisation/virtualisation.rst
plbossart: this is really confusing. pm_runtime is all about devices, and here we have vhost/guest devices. I really have no idea how pm_runtime can be synchronized unless there's a clear protocol and refcounting. 

BTW why this would this be audio specific, we can't be the only ones enabling pm_runtime?

developer_guides/virtualisation/virtualisation.rst
lgirdwood: host owns all DSP PM, and will only suspend if all guests are also suspended (guests send the PM IPC - host should also delete guest pipelines at this point). This means that on guest resume code flow is exactly the same as host.

developer_guides/virtualisation/virtualisation.rst
lyakh: no, only the in-kernel vhost.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood I'm not sure we need the mailbox-doorbell concept here. In my understanding the mailbox-doorbell interface is symmetric, at least that's (at least conceptually) what it is like with SOF. VirtQueues OTOH are asymmetric. In fact only one side can be the initiator. That's why we have to pre-queue position update buffers from the guest. The opposite direction is always a response.

lyakh: @plbossart yes, SOF-defined

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart is it worth mentioning that? That isn't really VirtIO specific, that's generic

developer_guides/virtualisation/virtualisation.rst
lgirdwood: Yes, runtime PM ref counting should manage this on bost host and guest. host should never enter D3 is guest is in D0.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: The host should at least have an idea of the pipelines and components that belong to each guest. This way it can delete them in FW if guest crashes etc.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: we do, we need to explain that we are virtualising a mailbox and doorbell. we are not virtualising PCMs or kcontrols as some people may expect.

lyakh: ok, got it

developer_guides/virtualisation/virtualisation.rst
lyakh: I asked on the official QEMU IRC, got no replies. That's certainly not a 100% reliable indication, but usually they're responsive, and they did try to help with this one too initially. But yes, we should ask on the mailing list.

developer_guides/virtualisation/virtualisation.rst
lyakh: yep, will do

developer_guides/virtualisation/virtualisation.rst
lyakh: well, I wasn't sure whether I had to put this here. It's already rather low level, but looks like I have to explain more.

developer_guides/virtualisation/virtualisation.rst
lyakh: this is the only configuration I've tested so far. I think it should also work with a dedicated part-pipeline on the host, but this hasn't been attempted yet.

developer_guides/virtualisation/virtualisation.rst
lyakh: kind of, yes. Both those component IDs belong to the same range, but the "guest ID" is an ID of a virtual DAI, that is only present on the guest, and that is then replaced with a component ID from the host part of the pipeline, to which the guest part is attached

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood the host has guest component ID ranges, is that enough?

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood you're proposing to explicitly synchronise the host runtime PM to the guest (runtime or system-level) PM? This is currently done semi-explicitly by the choice of auto-suspend delays. Is that enough at least for now?

developer_guides/virtualisation/virtualisation.rst
lyakh: There is a check in topology tools, I think, checking that one of the sides of any connection is a buffer

developer_guides/virtualisation/virtualisation.rst
lyakh: Would this be ok if I keep references to mixer / demux but explain, that this isn't a requirement, but just the currently used configuration? At least I don't think I have any explicit checks or dependencies for that in the code. I only care about linking component IDs

developer_guides/virtualisation/virtualisation.rst
lyakh: "running in a VM" means a guest driver, yes

lyakh: ...or rather the SOF driver, running on the guest

developer_guides/virtualisation/virtualisation.rst
lyakh: hm, the SOF driver on the guest only knows about components, that are parts of its topology. But in principle if it were to send an IPC with a "wrong" component ID, it would probably also be executed. We can add checks in the future, but that's of course an overhead. But if we manage to find a way to reliably identify guests, we will be sure, that every guest is trust-worthy, so, we shouldn't need any such checks.

developer_guides/virtualisation/virtualisation.rst
lyakh: only vhost, only in-kernel, no user-space. QEMU only establishes a connection between the guest and the vhost driver, but it doesn't actually serve the queues.

developer_guides/virtualisation/virtualisation.rst
lyakh: it's a dedicated virtualisation ID from RedHat

developer_guides/virtualisation/virtualisation.rst
lgirdwood: yes, but it needs to tell FW to delete the pipeline and components if needed. host may need to keep a list of all guest comps/pipes for each guest.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: no, I mean guests will increment / decrement host PM count by sending specific IPCs (as a result of userspace actions), these actions should be the same ones used bu host driver to inc/dec PM counts.

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart you mean various audio processing modules on the DSP? Why cannot guests use them directly? Is this impossible due to some principal restriction or is this something, that we shouldn't allow due to some concerns?

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart buffers - yes. Interrupts - you mean position updates?

developer_guides/virtualisation/virtualisation.rst
lyakh: Audio routing. We are using the same dynamic audio routing API as the one used with mixer kcontrols. I think.

developer_guides/virtualisation/virtualisation.rst
lyakh: We don't have 0-copy yet, so I'm trying not to speak too much about it. I only mention it to explain our eventual intention and to indicate where we already expect differences. But yes, once we get 0-copy, we might also keep the current implementation as a fall-back.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: @lyakh modules will reside on the host and it's up to host to decide whether to load them into DSP FW upon a guest request (since they consume DSP mem and cycles). We dont have IPC infrastructure to do this atm though so does not need implemented yet in virtio code.

developer_guides/virtualisation/virtualisation.rst
plbossart: it's really not great to explain as follows:

guest can connect to foobar but foobar is not a requirement and other connections are possible

You may want to explicitly state what the guest can connect to, so that there's no ambiguity for the reader/reviewer.

developer_guides/virtualisation/virtualisation.rst
plbossart: what I meant is that we don't necessarily want to use ONLY zero-copy in the end, it's useful to keep the mechanism with copies to e.g. debug issues, proceed with faster integration on specific platforms, etc.

You present the solution as exclusive, and I believe keeping the copy-based solution can be useful at times.

developer_guides/virtualisation/virtualisation.rst
plbossart: I didn't see any references to audio routes in that link. I guess you are probably talking about DAPM audio routes. Audio routing is a very vague and broad concept, you are using it in a very specific context that needs to be explicit.

developer_guides/virtualisation/virtualisation.rst
plbossart: then make it clear!

developer_guides/virtualisation/virtualisation.rst
lyakh: No replies on the mailing list either. @lgirdwood what exactly do you mean by leaving room in *any* init flow? We already pass a guest ID / topology file name with "request topology" - is that enough or you mean something different?

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart but it isn't just one of many possibilities, I think there are 2 options: BE sharing with the host or other guests and dedicated BEs. When sharing you use mixer / demux, when dedicated you use more or less anything else?

developer_guides/virtualisation/virtualisation.rst
lgirdwood: @lyakh ask Eddie

lyakh: @lgirdwood ok, will do. But I'm wondering: if the topology simply resided in the guest's filesystem, we wouldn't need any authentication? Could you remind me again why this wasn't acceptable, please?

developer_guides/virtualisation/virtualisation.rst
lyakh: mmh, I'm not an expert, but looking at dapm.rst and dpcm.rst it seems to me, that DAPM is mostly about Power Management. It uses audio routing information to make PM decisions, but to activate and deactivate guest audio streaming we use DPCM, which then uses DAPM to manage widget power states.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: guests are not trusted or safety critical certified.

lyakh: ok, yes, my mistake, letting a guest send its own topology would be in no way better than letting it choose which topology it should receive, in fact even worse.

lyakh:
I think this version addresses all comments so far with 2 exceptions:
1. topology tear down
2. host runtime PM triggered from guest runtime PM callbacks
I'd first implement them in the code, and then add them in the document. Would that be ok? I've also added an image from @lgirdwood 's slides - thanks.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: Please put 
```
This document describes the Linux driver architecture to virtualise SOF DSP firmware features across multiple guest VMs.

It describes a 

1) SOF VirtIO (Virtual Frontend) driver to be run on a guest virtual machine.
2) SOF vhost (Virtual Backend) driver to be run on the host.

This driver architecture allows full SOF audio use cases to reside on either guest VM or host. It also allows non audio DSP firmware features to also be virtualised on guests.
```

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,325 @@
...
+is an Android VM running in a ChromeOS environment.
+
+2. SOF VirtIO design
+********************
+
+The SOF VirtIO design virtualises the IPC command and audio data flows between
+host and guests. It re-uses the SOF-defined IPC API, also used by the native SOF
+driver to communicate with the SOF firmware, running on the audio DSP. This
+means, that the virtualisation is "applied" between the SOF driver and the DSP.
+Typically IPC message exchange between the driver and the DSP takes place over a
```
lgirdwood: new paragraph, also explain IPC.

developer_guides/virtualisation/virtualisation.rst
plbossart: yes, this description is accurate and should be added

lyakh:
> I would still add more pictures from the slides, especially at the introduction and also add more section headers. The document also appears to be a description of your implementation rather than a high level arch/design doc.

@lgirdwood that's what I indeed tried to do. We were discussing my patch series and its design and implementation were unclear, and replying in the PR thread became meaningless - you couldn't see old questions any more, so they got repeated. Then I proposed to write a description of the *design and implementation*. At some point it became more of a user-facing document, than a help to reviewers and developers, which isn't bad, but maybe we need both? One document, describing potentially complex or controversial technical details, another one with a high-level user-oriented description?

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart it's already there in the latest version

lgirdwood:
> > I would still add more pictures from the slides, especially at the introduction and also add more section headers. The document also appears to be a description of your implementation rather than a high level arch/design doc.
> 
> @lgirdwood that's what I indeed tried to do. We were discussing my patch series and its design and implementation were unclear, and replying in the PR thread became meaningless - you couldn't see old questions any more, so they got repeated. Then I proposed to write a description of the _design and implementation_. At some point it became more of a user-facing document, than a help to reviewers and developers, which isn't bad, but maybe we need both? One document, describing potentially complex or controversial technical details, another one with a high-level user-oriented description?

The intention here was to make it easier to understand all your code, so keeping it high level is key, at least for the bulk of the design document otherwise it just turns into a English version of your code which can be difficult to follow.

You can still keep the implementation section, but just make it easier to follow if describing sequences etc, also bitmask descriptions are not very interesting.

lyakh:
> You can still keep the implementation section, but just make it easier to follow if describing sequences etc, also bitmask descriptions are not very interesting.

@lgirdwood What about all the specific references to the code? function, macro, struct field names? All that certainly doesn't belong to a high-level design document.

lgirdwood:
> > You can still keep the implementation section, but just make it easier to follow if describing sequences etc, also bitmask descriptions are not very interesting.
> 
> @lgirdwood What about all the specific references to the code? function, macro, struct field names? All that certainly doesn't belong to a high-level design document.

Yes, this would be in the implementation sections rather than the high level sections.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lgirdwood I've got some replies from Eddie and other guys, he involved in the conversation. Basically, I think, what they are proposing, is make per-guest (audio) configuration decision based on guest's instance itself, i.e. eventually on guest's VM disk image. This of course asks for some host user-space support. On the host you would have a configuration file like
```
image: /path/to/vm/images/vm0
audio-topology: /path/to/vm/audio/topology/vm0
```
then SOF support code in QEMU will have to know to send this information to the SOF vhost driver during initialisation. Then the guest SOF virtio driver will just ask for its topology and the SOF vhost driver will already know which one belongs to this VM. Would this be acceptable?

lgirdwood: Yep, good for me if Eddie approves.

plbossart: > then SOF support code in QEMU will have to know to send this information to the SOF vhost driver during initialisation. 

What is 'this information'? and how would it be sent? and how would the host do a match?

lyakh: @plbossart "QEMU support code" means the SOF patch for QEMU. It opens the /dev/vhost-dsp device, sends ioctl()s to it, configures virtuql queues etc.
"This information" means, once qemu starts with the disk image vm0, that code will check the table in /etc/qemu/driver-conf.d/sof.conf (the path is invented) and send an `ioctl(GUEST_ID, &(struct guest_id){.id = "vm0"});` to the SOF vhost driver.

lyakh:
@lgirdwood @plbossart could you review again, please

developer_guides/virtualisation/virtualisation.rst
lgirdwood: not seeing this, have you pushed ?

lyakh: @lgirdwood yes, it's there. It now has 4 entries instead of 2, maybe that's why you didn't notice it?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+implement a VirtIO host: in the user space and in the kernel. If implemented in
+the user space, it is managed by QEMU. In the kernel space VirtIO host drivers
+use the Linux vhost API.
+
+It is also worth to note, that VirtIO guest drivers reside in respective driver
+subdirectories in the Linux kernel, e.g. drivers/net, drivers/block etc. On the
+contrary, all vhost drivers are located under drivers/vhost. That directory also
+contains vhost.c and vhost.h which are used by all vhost drivers.
+
+This document describes the Linux driver architecture to virtualise SOF DSP
```
lgirdwood: This should be at start of the document. all the way down to....

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+
+1. SOF VirtIO (Virtual Frontend) driver to be run on a guest virtual machine.
+2. SOF vhost (Virtual Backend) driver to be run on the host.
+3. SOF audio topology for both host and guest.
+4. SOF supporting code in QEMU.
+
+This driver architecture allows full SOF audio use cases to reside on either
+guest VM or host. It also allows non audio DSP firmware features to also be
+virtualised on guests.
+
```
lgirdwood: to here.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+********************
+
+The SOF_ VirtIO design virtualises the IPC (Inter-Process Communication) command
+and audio data flows between host and guests. It re-uses the SOF-defined IPC
+API, also used by the native SOF driver to communicate with the SOF firmware,
+running on the audio DSP. This means, that the virtualisation is "applied"
+between the SOF driver and the DSP.
+
+Typically IPC message exchange between the driver and the DSP takes place over a
+mailbox and a doorbell. A mailbox is a PCI device memory area, accessible from
```
lgirdwood: "A mailbox is a shared memory region between DSP and host."

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+both the CPU and the DSP. It is used for IPC message payload. A doorbell is a
+set of PCI registers, used to signal the DSP and the CPU about a message status
```
lgirdwood: Doorbell is asynchronous signalling hardware between DSP and host.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: The goal of the SOF virtualisation is to support differentiation in guest audio capabilities and implement low latency audio....

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+
+When configured to support virtualised audio, the host topology additionally
+provides connection points for guest audio pipelines. Guest playback pipelines
+might for example connect to mixer and capture pipelines to demultiplexer
+components. This way both the host and possibly multiple guests can share parts
+of the audio topology. It is also possible to dedicate certain audio pipelines
+to specific guests only. In such cases guest pipeline fragments can be connected
+to other component types, e.g. to a volume control. We'll use the term
+"connection points" throughout this document to refer to these locations in the
+host audio topology.
```
lgirdwood: Can this be guest/host connections be shown in a diagram.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+currently performing audio playback. To achieve this, for each such guest
+connection point we add a virtual PCM as a widget of type "aif_in" or "aif_out"
+for playback and capture respectively, and a virtual 0-size buffer. They are
+only visible on the host and are not sent to the DSP. When processed by the
+ASoC topology parser, those virtual PCM widgets create front-end DAI links and
+ASoC PCM runtime contexts. These will later be used for guest audio stream
+management.
+
+When running in a VM, the SOF driver also uses ALSA topology, but in this case
+the topology should have no relation to the host hardware. The guest audio
```
lgirdwood: the guest topology should have....

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+When QEMU starts a guest instance, it detects the above misc device and
+creates a PCI device for the guest with dedicated vendor and device IDs. All
+vhost PCI devices use the Qumranet / RedHat vendor ID and are allocated a
+device ID from a range, specially donated by RedHat for this purpose.
+
+The SOF VirtIO driver on the guest registers support for that PCI device and
+its .probe() method is called.
+
+During the SOF VirtIO driver probing, the QEMU SOF code reports to the vhost
+driver, which VM image is being instantiated. This is used as a ewliable guest
```
lgirdwood: ewliable ?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+The SOF driver on the guest performs partial initialisation, it skips any
+steps, involving communicating with the actual DSP, e.g. sending the firmware,
+booting it, initialisation the tracing interface.
+
+The VirtIO driver on the guest uses a new SOF_IPC_TPLG_VFE_GET VirtIO-specific
+IPC message to request the topology from the host. The host reads the file,
+corresponding to the guest ID, obtained from the QEMU communication described
+above, from its filesystem, and sends it in multiple chunks to the guest in
+reply to that request.
+
```
lgirdwood: I think there are some slides showing the boot flow.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+the host IPC handler just forwards IPCs to the DSP, but in some cases it has to
+handle or modify IPC contents locally. One such example is linking the guest and
+host topology fragments together. As mentioned above, guest topology files
+contain a new token, referencing connection point components in the host
+topology. This value is then assigned to the .ref_comp_id field of struct
+sof_ipc_comp_config during topology file parsing and sent by the guest to the
+host. The host then uses that value to modify SOF_IPC_TPLG_COMP_CONNECT
+connection IPC messages, involving connection points, from the guest before
+sending them to the DSP.
+
```
lgirdwood: So this is a complex sequence and would be best represented with a diagram.

developer_guides/virtualisation/virtualisation.rst
lgirdwood: ack, ref counting on host for all guests will manage this in general purpose way.

developer_guides/virtualisation/virtualisation.rst
lyakh: @lrgirdwo Should I then throw away the explanation of VirtIO in Linux? It would be wrong to have that after this text. Or we can add a "Preliminary information" subsection?

lgirdwood: it can be kept, the main point of this document is to describe SOF virtualisation and not virto, so SOF stuff would be the first paragraph.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+Guest topology files contain virtual DAI components of type "dai_out" and
+"dai_in." Their data tuples contain a new token of type SOF_TKN_COMP_REF, which
+contains an ID of the connection point, to which this pipeline should be
+attached.
+
+After obtaining the topology, the guest SOF instance uses a second new
+SOF_IPC_TPLG_VFE_COMP_ID VirtIO-specific IPC message to request its allocated
+component ID base. The host maintains a list of guest component ID ranges to
+guarantee, that component IDs stay unique regardless of the order in which
+guests are brought up and down.
```
plbossart: can you clarify this: around Line 178 you write "volume component ID as seen by the respective guest. Therefore it should be exactly the same ID, as the one used by the firmware."

so how does the notion of guess-specific component ID base work?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+host. The host then uses that value to modify SOF_IPC_TPLG_COMP_CONNECT
+connection IPC messages, involving connection points, from the guest before
+sending them to the DSP.
+
+3.2. Streaming
+--------------
+
+Guest audio streaming is mostly transparent for the host. The host audio
+subsystem doesn't get involved with most guest streaming or kcontrol operations
+because the virtualisation is applied at the DSP level, not at the audio level.
```
plbossart: i don't get what is 'DSP level' and what is 'audio level'?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+particular care, because in the present state, just calling
+
+|	soc_dpcm_runtime_update();
+|	snd_soc_runtime_deactivate();
+
+doesn't deactivate the pipeline and therefore doesn't allow runtime PM to
+suspend the interface, which then breaks following activation attempts. The
+current SOF VirtIO implementation contains a fix for that, which has to be
+upstreamed along with the rest of the ALSA core virtualisation modifications
+and extensions.
```
plbossart: runtime_pm on the host only works with open() and close() operations. It doesn't matter if there is a trig_start/stop, the refcount is only based on open/close. I see danger here if we use different concepts for virtualization. it should be when the PCM stream is closed that pm_runtime can kick in (and with a N s delay).

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+
+3.3. Deactivation
+-----------------
+
+When the guest isn't actively using audio, we want to allow the host to
+runtime suspend the DSP. That usually means switching the DSP off. Next time the
+guest wants to activate audio it will have to perform initialisation again,
+including sending its topology to the host. This happens automatically as long
```
plbossart: the topology file is provided by the host to the guest, I guess you must mean using IPC to configure the graph on the DSP during the topology parsing?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+It describes a 
+
+1. SOF VirtIO (Virtual Frontend) driver to be run on a guest virtual machine.
+2. SOF vhost (Virtual Backend) driver to be run on the host.
+3. SOF audio topology for both host and guest.
+4. SOF supporting code in QEMU.
+
+This driver architecture allows full SOF audio use cases to reside on either
+guest VM or host. It also allows non audio DSP firmware features to also be
+virtualised on guests.
```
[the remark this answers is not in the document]
lyakh: This section has been contributed by @lgirdwood, personally I would just mention a possibility of non-audio components briefly once, or maybe leave it out completely.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+of standard and mostly empty device node and virtual queue management
+operations.
+
+2. SOF VirtIO design
+********************
+
+The SOF_ VirtIO design virtualises the IPC (Inter-Process Communication) command
+and audio data flows between host and guests. It re-uses the SOF-defined IPC
+API, also used by the native SOF driver to communicate with the SOF firmware,
+running on the audio DSP. This means, that the virtualisation is "applied"
```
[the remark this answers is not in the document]
lyakh: well, isn't it a rather standard notion for virtualisation? You can "apply" virtualisation at different levels: you can virtualise hardware, you can virtualise an OS, you can virtualise library APIs...

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+host audio topology.
+
+The ALSA / ASoC / DAPM subsystems on the host are unaware of any guests, but
+they should be aware of any streaming, taking place on host parts of VM
+pipelines for configuration and power management. That means, e.g. if a guest
+initialises playback, the ASoC subsystem on the host should be aware, that the
+pipeline, joining the respective connection point and the used codec is
+currently performing audio playback. To achieve this, for each such guest
+connection point we add a virtual PCM as a widget of type "aif_in" or "aif_out"
+for playback and capture respectively, and a virtual 0-size buffer. They are
```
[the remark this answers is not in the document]
lyakh: @plbossart I think it's ALSA topology tools, that mandate one side of every topology connection to be a buffer. So, on the host in principle I just need
```
[ fake / virtual PCM ]
[ connection point, e.g. a mixer]
...
[codec]
```
where the fake / virtual PCM is for DPCM route updating on the host and is not sent to the DSP. But such a pipeline is invalid, there *must* be a buffer between that PcM and the mixer. And since we don't really need that buffer and it isn't sent to the DSP, we make it 0-byte large.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,349 @@
...
+2.3. DPCM audio routing
+-----------------------
+
+In the most trivial case when a user-space application opens an audio interface,
+there is a unique sequence of audio components, that will comprise a DPCM audio
+route, involved in this operation. Such a sequence is called an audio pipeline.
+However, modern audio hardware often contains more complex audio graphs, also
+sometimes requiring dynamic re-routing. Support for such configurations is
+provided by the ASoC DPCM_ API. We also use this API to activate and deactivate
+guest audio interfaces.
```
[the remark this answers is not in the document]
lyakh: Perhaps... Is a "normal" PCM a DPCM FE?

developer_guides/virtualisation/virtualisation.rst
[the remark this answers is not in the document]
plbossart: We don't have a solution at the moment and if we did it wouldn't rely on ASoC....

developer_guides/virtualisation/virtualisation.rst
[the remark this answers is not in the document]
plbossart: It's worth adding that this is an ALSA topology restriction, I thought it was a virtio one.

developer_guides/virtualisation/virtualisation.rst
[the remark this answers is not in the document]
plbossart: in general yes. DPCM BEs are typically marked as .no_pcm = 1;

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart this isn't about runtime PM, it's about audio runtimes / pipelines. It affects the runtime PM but only indirectly. As I mentioned in my today's comment to https://github.com/thesofproject/linux/pull/1501 earlier today, I now have a version of runtime PM, which works synchronously to the guest rt-pm. A description of runtime-PM in this document is mostly described in section 3.3 below. If we agree to use the new SOF VirtIO runtime-PM concept, I'll update that chapter too.

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart I mean booting the DSP for each runtime-resume. When the DSP boots, the driver has to re-send the topology to it. This is what is done now at least in some configurations. Here we have to do the same but we need to send all topology fragments - from the host and any guests.

developer_guides/virtualisation/virtualisation.rst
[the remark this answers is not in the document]
lyakh: re-reading that text I found the first sentence rather unclear, so I improved it. But this part I actually find rather logical: first a sentence describes, that in the native case the driver uses IPC to talk to the DSP, then the following sentence says, that the virtualisation is applied right there - *between* the driver and the DSP. But maybe it's the first sentence of the section that made that part confusing. Hopefully after I improve it, it will make the following text clearer too.

developer_guides/virtualisation/virtualisation.rst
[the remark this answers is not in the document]
lyakh: Ok, I add this to 2.2 above, where I describe these PCMs

developer_guides/virtualisation/virtualisation.rst
plbossart: the host sends the topology binary to the guest.
The guest will issue a set of IPC to create the topology on the DSP.

Can you detail the steps on resume, e.g. can we assume the guest has kept a copy of the topology file and only needs to recreate the graph? or has the guest lost all context?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+
+2. SOF VirtIO design
+********************
+
+The SOF_ VirtIO design implements an interface to transport IPC (Inter-Process
+Communication) commands and audio data between host and guests. It re-uses the
+SOF-defined IPC API, also used by the native SOF driver to communicate with the
+SOF firmware, running on the audio DSP. This means, that the virtualisation is
+"applied" between the SOF driver and the DSP, i.e. IPC messages that the guest
+SOF driver would send to the DSP are instead sent to the guest, which then in
```
plbossart: error: sent to the host.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+operations.
+
+2. SOF VirtIO design
+********************
+
+The SOF_ VirtIO design implements an interface to transport IPC (Inter-Process
+Communication) commands and audio data between host and guests. It re-uses the
+SOF-defined IPC API, also used by the native SOF driver to communicate with the
+SOF firmware, running on the audio DSP. This means, that the virtualisation is
+"applied" between the SOF driver and the DSP, i.e. IPC messages that the guest
```
plbossart: virtualization mechanisms are inserted between the SOF guest driver and the DSP hardware.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+support them. For the most part IPCs from guests are forwarded without change by
+the host driver to the DSP, only some of them are intercepted on the host. Some
+IPC messages trigger additional actions on the host before being forwarded to
+the DSP, others aren't sent to the DSP at all. See examples below for more
+details.
+
+2.1. Communication channels
+---------------------------
+
+The goal of the SOF virtualisation is to support differentiation in guest audio
```
plbossart: I understand what you mean by differentiation but that's an Intel buzzword.

maybe break in two objectives
- first goal is to enable streaming to/from audio endpoints, with the lowest possible overhead
- second goal is to allow a guest to control some of the processing on the DSP, while isolating it from other parts of the graph owned by the host or other VMs.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+and down.
+
+Next the guest proceeds by sending the parsed topology over IPC to the host,
+which is then forwarded to the DSP firmware. The guest uses the .send_msg()
+callback to forward any IPC, including topology related, with no processing at
+all, to the host. The only exception is blocking IPCs when the guest is
+resuming, while the host doesn't need to be resumed (see section
+`3.3. Deactivation`_ on `reset count`_ for details).
+
+The host receives the guest topology over the IPC virtual queue. In most cases
```
plbossart: the host provides the topology file to the guest
the guest sends topology/configuration IPCs to the host.

which of the two cases does line 255 describe?

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+the host IPC handler just forwards IPCs to the DSP, but in some cases it has to
+handle or modify IPC contents locally. One such example is linking the guest and
+host topology fragments together. As mentioned above, guest topology files
+contain a new token, referencing connection point components in the host
+topology. This value is then assigned to the .ref_comp_id field of struct
+sof_ipc_comp_config during topology file parsing and sent by the guest to the
+host. The host then uses that value to modify SOF_IPC_TPLG_COMP_CONNECT
+connection IPC messages, involving connection points, from the guest before
+sending them to the DSP.
```
plbossart: maybe a stupid question, but couldn't we make sure that the topology sent by the host to the guest contains a set of unique component IDs, so that the guest never ever conflicts with other topologies, and that way we would remove the need for this translation layer?

We do want to uniquify components but do we really need the component base and translation back when the topology could implement unique IDs from day 1?

If you enable a range-based check in the host driver to make sure the IPCs sent by the host only target the components it instantiated and controls that would be a different story, but you didn't mention it here.

developer_guides/virtualisation/virtualisation.rst
```diff
@@ -0,0 +1,360 @@
...
+subsystem doesn't get involved with most guest streaming or kcontrol operations
+because the virtualisation is applied at the SOF hardware driver level, not at
+the ASoC API level. All those operations are completely processed on the guest
+and only resulting DSP IPCs are forwarded to the host. The only exceptions are
+beginning and ending of the streaming, at which times the ASoC subsystem on the
+host has to be informed, that a certain pipeline is becoming active or inactive
+respectively.
+
+The SOF vhost driver uses the STREAM_PCM_PARAMS IPC message from the guest to
+allocate an ALSA PCM runtime object and configure the HDA hardware. If zero-copy
```
plbossart: You should remove references to HDA, this could work on NXP hardware as well.

developer_guides/virtualisation/virtualisation.rst
lyakh: mmh, between the host driver and the DSP? As far as I understand what we are saying is, that the normal case is no virtualisation. Just a (host) SOF driver and a DSP, communicating over IPC. Now we take that interface and bring the "top" part to the guest. So, we "insert" those mechanisms into the original (host) driver?

developer_guides/virtualisation/virtualisation.rst
lyakh: the latter. I'll try to make it clearer

developer_guides/virtualisation/virtualisation.rst
lyakh: @plbossart Component IDs are assigned during topology parsing, right? Guest topology fragments are parsed on guests, the host doesn't look into them at all (prior to sending). The host just reads in files from the filesystem and sends them out to guests. Besides component IDs aren't even present in topology files, there are probably even no fields for them there. So, we would have to change a lot for that - add processing on the host before sending, and change processing on the guest, because the data, it receives now isn't the standard topology data any more. So, no, don't think it's a good idea.

Topology IDs are unique, that's why we use their ranges. Translation is a different thing. It isn't used for all IDs. It's only used for connection points. Guests send their complete pipelines to the host. (See image). So, e.g. such a guest would send
```
[PCM]
[buffer]
[backend]
```
but the backend is "fake" / "virtual." What actually should happen is that the above PCM and buffer are connected to, say, a mixer on the host pipeline. So, when the guest sends the "backend" widget to the host, it will be dropped, instead of being sent to the DSP. And when the guest sends "connect buffer ID x to backend ID y" - *this it* where the translation happens. The host then replaces ID y with ID z of the mixer.

I don't have such a check yet, but that would also be a 1-2 line addition. All the data is already there. But again, that has nothing to do with connection point ID translation.

developer_guides/virtualisation/virtualisation.rst
plbossart: oh, I see what you're saying now.

I don't really have a better wording so let's keep it as is.

developer_guides/virtualisation/virtualisation.rst
plbossart: Ah, I assumed the component IDs were part of the topology, but indeed if they are assigned dynamically while parsing then that's a wrong assumption. I may have learned something today :-)

@ranj063 can you confirm @lyakh 's point?

deb-intel:
@lgirdwood Sounds good. I will merge it and publish all merged docs before my EOD.

deb-intel:
@plbossart Thanks for the updates.

deb-intel:
@lyakh @lgirdwood @plbossart 
Note: I made numerous updates to this PR, including file name changes (Intel compliance). Refer to PR 163 comments for details.

PR #169, opened 2020-01-28
Grammatical edits to Algorithms doc

deb-intel:

### DIFF
--- a/algos/index.rst
+++ b/algos/index.rst
@@ -3,7 +3,7 @@
 Supplied Processing Algorithms
 ##############################
 
-SOF contains several permissively licensed and royalty free audio processing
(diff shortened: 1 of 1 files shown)


deb-intel:
@singalsu Please review my edits in the description section, especially "Mixer."

singalsu:
The mixer description can be improved "Sum of unity gain and multiple pipeline saturation to a single output" -> "Sums with unity gain and saturation source buffers of multiple pipelines to a single output sink buffer". The processing description "unity gain and saturation" could be omitted as too detailed if you prefer.

I always do some mistakes with grammar so also that may need some adjust!

deb-intel:
@singalsu Great! Thanks!

PR #188, opened 2020-02-19
Tighten nav bar headings: platform, algorithms, dev guides

deb-intel:


deb-intel:
The navigation bar contained many sub-headings. These sub-headings have been moved so they no longer appear on the main level.

PR #200, opened 2020-03-18
comp: removed doc duplicated in dox

mmaka1:
All duplicated descriptions removed. More references to
dox-generated part added. The latter takes precedence
and is more up to date as generated from the source code.

Removed static diagrams that quickly go out-of-sync with
the implementation if they do not bring much value.


deb-intel:
@mmaka1 Are we certain that developer_guides/firmware/components/component-api.rst is not referenced anywhere else on the SOF website? Just checking . . . .

PR #202, opened 2020-03-20
probes: add probes documentation

abonislawski:
Simple getting started with probes

Signed-off-by: Adrian Bonislawski <adrian.bonislawski@linux.intel.com>


abonislawski:
@lgirdwood we are done in FW but usage depends on kernel implementation which is not ready, so it will be added later

PR #203, opened 2020-03-20
[RFC] add mux/demux documentation

juimonen:
Signed-off-by: Jaska Uimonen <jaska.uimonen@intel.com>


algos/demux/demux.rst~
```diff
@@ -0,0 +1,78 @@
+.. _demux:
+
+Muliplexer/Demultiplexer
+########################
```
lgirdwood: This looks like a duplicate of the previous file ?

juimonen: oh, emacs backup, sorry will remove.

juimonen:
@lgirdwood updated. 
- removed emacs backup file
- removed white space
- fixed a typo in header

This is now depending on: 
- https://github.com/thesofproject/sof/pull/2594 
- https://github.com/thesofproject/sof/pull/2600.

I also talk about python blob generator: https://github.com/thesofproject/sof/pull/2536, which people seem not to like, but it is the only "generic" tool we would have to generate blobs... But I can remove this if no traction. So the python tool could be used to generate dc blocker, mux/demux, or even simple eq blobs.

And now when I'm on it: we should really do the optimization for demux @Keyon was talking about in https://github.com/thesofproject/sof/issues/2533.

Then the demux would be in pretty good shape, so this here looks like mux todo list :)

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+input to output stream. As ASoC/SOF audio stream can have up to 8
+audio channels, a stream-to-stream specific 8 x 8 routing matrix
+defines the channel mapping from input to output. Because every stream
+is fully configurable we have a matrix for all multiplexer input
+streams or all demultiplexer output streams. 8 x 8 binary matrix takes up
+to 64 bits and is controlled with eight unsigned char values.
+
+.. figure:: images/mux.png
+
+   Example of multiplexer configuration matrises with 2 input streams.
```
ranj063: matrises -> matrices?

ranj063: or maybe just matrix?

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+   In this artificial Mux example Input stream 1’s channel 1 is copied
+   to both output channels, Input stream 2’s single channel is copied
+   also to output channel 1.
+
+.. figure:: images/demux.png
+
+   Example of demultiplexer configuration matrises with 2 output streams.
```
ranj063: ditto

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+   output streams)
+
+Using a routing matrix means also that this component can leave out
+channels or mix multiple channels into one.
+
+Topology
+========
+
+Previous figures show that the routing matrix is difficult to
+parametrize to be easily understandable. As it is send to firmware
```
ranj063: send->sent?

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+with 64 bits, it is quite tedious to easily see from hexadecimal or
+integer values the binary routings. SOF topology m4 macros have helpers
+to "visualize" the matrix for easier configuration.
+
+Example from pipe-volume-demux-playback.m4 shows how to define 2
+routing matrises and a demux component:
```
ranj063: matrices

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+		W_MUXDEMUX(0, 1, PIPELINE_FORMAT, 2, 2, LIST(`	       ', "DEMUX"))
+
+In the above example you can see the routing matrises have only
+"diagonal" 1's, which means in practice that input stream's channels
+are copied to corresponding output streams channels.
+
+Alsa control
+============
+
+Multiplexer configuration is loaded in kernel/firmware boot as part of Alsa
```
ranj063: ALSA

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+binary control in topology, but can be also controlled through alsa
```
ranj063: ALSA all caps

algos/demux/demux.rst
juimonen: yes

algos/demux/demux.rst
juimonen: yes

algos/demux/demux.rst
juimonen: yes

algos/demux/demux.rst
juimonen: yes

algos/demux/demux.rst
juimonen: ok

algos/demux/demux.rst
juimonen: ok

juimonen:
@ranj063 thanks for comments! updated now, hopefully addressed all comments...

algos/demux/demux.rst
```diff
@@ -0,0 +1,138 @@
...
+is fully configurable we have a matrix for all multiplexer input
+streams or all demultiplexer output streams. 8 x 8 binary matrix takes up
+to 64 bits and is controlled with eight unsigned char values.
+
+.. figure:: images/mux.png
+
+   Example of multiplexer configuration matrices with 2 input streams.
+   In this artificial Mux example Input stream 1’s channel 1 is copied
+   to both output channels, Input stream 2’s single channel is copied
+   also to output channel 1.
```
libinyang, line 38: @juimonen I'm a newer to mux/demux. And I find it's a little hard for me to understand to understand the picture. Why do we need set column 3, 5, 7 to 1? Shouldn't it be 0s although it doesn't take efforts here. Or we must set 1s in column 3, 5, 7?

keyonjie:
> @lgirdwood updated.
> 
> * removed emacs backup file
> * removed white space
> * fixed a typo in header
> 
> This is now depending on:
> 
> * [thesofproject/sof#2594](https://github.com/thesofproject/sof/pull/2594)
> * [thesofproject/sof#2600](https://github.com/thesofproject/sof/pull/2600).
> 
> I also talk about python blob generator: [thesofproject/sof#2536](https://github.com/thesofproject/sof/pull/2536), which people seem not to like, but it is the only "generic" tool we would have to generate blobs... But I can remove this if no traction. So the python tool could be used to generate dc blocker, mux/demux, or even simple eq blobs.
> 
> And now when I'm on it: we should really do the optimization for demux @Keyon was talking about in [thesofproject/sof#2533](https://github.com/thesofproject/sof/issues/2533).
> 
> Then the demux would be in pretty good shape, so this here looks like mux todo list :)

Hi @juimonen, thanks for the documentation! Only one suggestion that to remove the complicate matrix part if we are aligned to remove the "channel mixing" feature in the future. The good point is that nobody is using that mixing feature until now, so don't mention that could be wise choice, to make easier life not only for people reading this documentation, but also for people who might need to maintain/update the documentation later.

algos/demux/demux.rst
juimonen: @libinyang just think as the row as input channel, and 0/1 as if this row (channel) is put to column output channel. so for example first row (0,1,0,1,0,1,0,1) would mean that copy input channel 1 to output channels 2, 4, 6, 8. And so on for other channels. There's always 1 matrix per stream-to-stream mapping.

deb-intel:
@lgirdwood Thanks!

PR #211, opened 2020-04-03
substitutions.txt: add missing |JSL|

marc-hb:
Fixes commit f4ada42ce36c ("getting_started: Add build documenation for
ICL and JSL+") that added four "WARNING: Undefined substitution
referenced: JSL" and for literal "|JSL|" instances on
https://thesofproject.github.io/latest/getting_started/build-guide/build-from-scratch.html,
example:

>  Current target Intel platforms include: Bay Trail, Cherry Trail,
>  Haswell, Broadwell, Apollo Lake, Cannon Lake, Ice Lake and |JSL|.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #212, opened 2020-04-03
Makefile: add ../sof/doc/doxygen/ dependency

marc-hb:
Doxygen is not incremental but it is super fast in this project with
this configuration, so run it every time before we run sphinx. Totally
negligible compared to the minutes required by plantUML.jar.

Keep it optional not to add new dependencies. Now for users who opt-in,
this commit will keep showing the doxygen warnings on every incremental
build (unlike the .rst warnings unfortunately).

Print the magic cmake invocation so no one has to go and read the build
HOWTO to get (re)started.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,18 +28,35 @@ help:
 	@echo "   specify RELEASE=name to publish as a tagged release version"
 	@echo "   and placed in a version subfolder.  Requires repo merge permission."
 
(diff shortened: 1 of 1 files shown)


marc-hb:
The warnings are not new and unrelated. One of them will be fixed by PR #211

PR #214, opened 2020-04-04
travis: add after_script step to check warnings

marc-hb:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,11 @@ install:
 
 script:
     - cd .. && git clone https://github.com/thesofproject/sof.git && cd sof/doc && cmake . && make doc && cd - && cd sof-docs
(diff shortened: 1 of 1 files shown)


marc-hb:
https://docs.travis-ci.com/user/job-lifecycle#breaking-the-build
> The exit code of after_success, after_failure, after_script, after_deploy and subsequent stages do not affect the build result.

Yet it's red above.

I'm adding `after_script` instead of `script` so `deploy` still runs even when `after_script` fails. Correct?

There's apparently no intermediate, "yellow" state in Travis, it's either green or red.

marc-hb:
I have a newer and much better version of this, I will submit it after #225 and https://github.com/thesofproject/sof/pull/2773 are merged.

PR #215, opened 2020-04-08
Travis: switch sof/doc sub-build to Ninja

marc-hb:
Since commit a97756ab4307 was merged, Travis shows the following harmless but
misleading message:
```
 # To build doxygen APIs too run this first:
 #   cmake -GNinja -S ../sof/doc -B ../sof/doc
```
Example: https://travis-ci.org/github/thesofproject/sof-docs/builds/670800271

That's because the sof-docs documentation and sof-docs/Makefile now
expect a sof/doc/ninja.build. So switch sof-docs/.travis.yml away from
Make and to ninja.

In the future we'll probably want cmake -S sof/doc to use a separate
-B build directory. One thing at a time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,14 @@ python:
 
 before_install:
     - sudo apt-get update -qq
-    - sudo apt-get install doxygen make default-jre graphviz cmake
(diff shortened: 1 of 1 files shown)


marc-hb:
See also https://github.com/thesofproject/sof/pull/2741

PR #218, opened 2020-04-09
Much faster build + increase sphinx-build verbosity

marc-hb:
Expand commit messages below


lgirdwood:
@intelkevinputnam any comments ?

PR #219, opened 2020-04-13
Small update to conf.py file to avoid build errors

deb-intel:


conf.py
```diff
@@ -191,3 +191,9 @@ def setup(app):
 }
 breathe_default_project = "SOF Project"
 breathe_default_members = ('members', 'undoc-members', 'content-only')
+
+try:
+    if "tox" not in exclude_patterns:
+        exclude_patterns.append(".tox")
+except:
+    exclude_patterns = [".tox"]
```
marc-hb, line 199: How can an `exclude_patterns` be anything else than the constant defined at the top?

intelkevinputnam: This is code is generated by another tool, which cannot assume that the list exists. If you'd prefer '.tox' can be added to the definition above. More generally, the conf.py get's executed before Sphinx runs, so any kind of Python code can be included, including registering methods with Sphinx to change the functionality of Sphinx.

marc-hb: >  If you'd prefer '.tox' can be added to the definition above.

I don't have any preference, at least not yet, just ignorance and fear :-) I made a one-line change to this file recently (a7e77d460) and now that you say it is generated I'm afraid I didn't know what I was doing.

> This is code is generated by another tool, which cannot assume that the list exists.

The comment at the top says it was generated the very first time but it doesn't seem to say anything about generating any part of it again...?

intelkevinputnam: I'm not a big fan of the conf.py as it mixes concerns: setting global variables and executable code. I'd prefer a true settings file: either .ini or .json. So fear justified. ;-) The conf.py is created by sphinx-quickstart when you create a new project and is one of the few required files for Sphinx. The snippet with exclude patterns is generated by the GUI app I mentioned - Sphinx goes nuts trying to parse its own .rst docs in the .tox directory if you don't. That's why it adds it at the end, so as not to mess with other settings. Both happen exactly once, so any other changes are fine.

PR #236, opened 2020-04-22
build-with-docker: Remove -l option from xtensa-build-all command

crazoes:
The -l option has been removed from the script for over a year since
https://github.com/thesofproject/sof/commit/9840ecbbfe41ce0cc53b5d0215c64c9f87557d39

Signed-off-by: Shreeya Patel <shreeya.patel23498@gmail.com>

### DIFF
--- a/getting_started/build-guide/build-with-docker.rst
+++ b/getting_started/build-guide/build-with-docker.rst
@@ -110,7 +110,7 @@ Build the SOF binaries:
 .. code-block:: bash
 
    cd ~/work/sof/sof/
-   ./scripts/docker-run.sh ./scripts/xtensa-build-all.sh -l
+   ./scripts/docker-run.sh ./scripts/xtensa-build-all.sh
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


crazoes:
The option itself was removed from the script since https://github.com/thesofproject/sof/commit/9840ecbbfe41ce0cc53b5d0215c64c9f87557d39 (Jan 18, 2019). However, the travis script on the repo was updated much later https://github.com/thesofproject/sof/commit/ffe904c9f9013b87d2c163eab0ca03ef34560315 (Oct 24, 2019) which leaves me confused because we couldn't possibly have had a broke CI for over 10 months. 

/cc commit authors: @jajanusz @xiulipan if you can provide some more info?

crazoes:
Okay, I might've looked into this a bit too deep :sweat_smile: .. travis didn't fail because the script didn't act to unknown options until https://github.com/thesofproject/sof/commit/e73344a73658cb0b16f5381954c4f41f5ff42ca6#diff-cd4f9307e85fe568a0351fa0d3e7b421

xiulipan:
@crazoes Thanks for the fix! The local was used for some old case and it is not needed anymore.

PR #244, opened 2020-05-19
contributing: Add section on process for SOF FW ABI changes

kv2019i:
This is documenting the proposed new, refined ABI process as discussed here:
https://github.com/orgs/thesofproject/teams/sof-developers/discussions/35

Document revised process for ABI changes including a new step
for Request-for-Comments phase. This is intended to ensure
early engagement of interface changes across driver and firmware
teams, to avoid wasting effort when concerns are raised at
the very end of the process.

Add cross-links between kernel driver and firmware sections.

### DIFF
--- a/contribute/process/images/abiprocess.pu
+++ b/contribute/process/images/abiprocess.pu
@@ -1,0 +1,23 @@
+[*] --> NewABIChange: Firmware feature that requires ABI change
+
+NewABIChange --> RFC: Send RFC Pull Request\ncovering interface changes\n and rationale for the change
+
+note right of RFC : RFC stage is intended to\navoid wasted effort via early\n engagement with the ABI users
+
+RFC --> RFCApproved: a) Approve+1 from at least\none Driver and one FW Maintainer,\n b) Owners assigned for FW and driver impl
+
+NewABIChange --> FWImplementation: Fast path implementation\nonly when no driver impact
+
+RFCApproved --> FWImplementation: Implementation done, submit as non-RFC PR
+
+FWImplementation --> ABIClassification: Tag PR for ABI classifier
+
+ABIClassification --> ABIApproved: TSC member approval and\nABI MAJOR.MINOR classification done
+
+ABIApproved --> DriverPRCheck: If driver change is needed,\nwait until both sides ready for merge
+
+DriverPRCheck --> FWChangeMerged: No driver change:\nAfter review and validation ok, merge
+
+DriverPRCheck --> DrvChangeMerged: After review and\nvalidation ok, merge
+
+DrvChangeMerged --> FWChangeMerged: After review and\nvalidation ok, merge
--- a/contribute/index.rst
+++ b/contribute/index.rst
@@ -14,4 +14,5 @@ patches for code, documentation, tests, and more, directly to the project.
    dox-source-code.rst
    process/bug-tracking
    process/docbuild
+   process/abiprocess
    linux/development_tree
--- a/contribute/linux/development_tree.rst
+++ b/contribute/linux/development_tree.rst
@@ -43,10 +43,11 @@ change tracker:
 
 https://github.com/orgs/thesofproject/projects/2
 
+Process for firmware ABI changes is documented in :ref:`SOF_ABI_changes`.
+
 When the ABI is not backwards-compatible, the Pull Requests on the
 kernel side shall include code that deals with older firmware and
 topology files.
-
 
 Development branch
 ==================
@@ -84,6 +85,7 @@ as audio codecs, or bug fixes may be contributed by the community. SOF Linux
 maintainers will, on a regular basis (typically weekly), merge all upstream
 contributions into the SOF tree.
 
+.. _sof_drv_maintainer_list:
 
 Development flow
 ****************
--- a/contribute/process/abiprocess.rst
+++ b/contribute/process/abiprocess.rst
@@ -1,0 +1,60 @@
+.. _SOF_ABI_changes:
+
+SOF ABI Change Process
+######################
+
+SOF ABI Definitions
+*******************
+
+SOF ABI consists of public structs used in host-FW communication
+defined in:
+
+- src/include/kernel/
+- src/include/ipc/
+- src/include/user/
+
+SOF ABI versioning is defined in firmware source code
+documentation `FW Source kernel/abi.h`_
+
+.. _FW Source kernel/abi.h: https://github.com/thesofproject/sof/blob/master/src/include/kernel/abi.h#L8
+
+Change Process
+**************
+
+When a firmware change requires extending or modifying the public
+SOF ABI, developer needs to go through the ABI change process as defined
+in this section. It is responsibility of the developer to drive this
+process, contact the stakeholders, request reviews (and re-reviews when
+needed) and e.g. coordinate with the driver maintainers.
+
+The main steps of the process are depicted in the following
+state diagram:
+
+.. _ABI Change Tracker: https://github.com/orgs/thesofproject/projects/2
+
+The pull requests are classified in GitHub using the following
+official `ABI Change Tracker`_
+
+.. uml:: images/abiprocess.pu
+   :caption: ABI process state diagram
+
+When the ABI change is not backwards-compatible, the Pull Requests on the
+kernel side shall include code that deals with older firmware and
+topology files. See :ref:`development_tree` for kernel side
+documentation.
+
+ABI Change Approvers
+********************
+
+TSC
+---
+
+Approval from SOF  :ref:`tsc`
+member is needed for all ABI changes.
+
+SOF Driver
+----------
+
+Linux driver team approval for changes can be granted by
+any member of the SOF Linux driver maintainer team. Current
+list of members is maintained in :ref:`sof_drv_maintainer_list`


kv2019i:
Minor update to diagram, there was a bug in the state diagram, now fixed. I also added a note that for MAJOR version bump, kernel implementation is mandatory. Updated generated picture at:
https://github.com/orgs/thesofproject/teams/sof-developers/discussions/35/comments/7

contribute/process/images/abiprocess.pu
```diff
@@ -0,0 +1,25 @@
...
+
+RFC --> DriverImplementation: Start working on driver changes
+
+NewABIChange --> FWImplementation: Fast path implementation\nfor minor changes
+
+note left of FWImplementation : Fast path only when no impact to driver
+
+FWImplementation --> ABIClassification: Submit PR for classification
+
+ABIClassification -> DriverImplementation: Start working on driver changes
```
lyakh, line 17: Why are there two paths marked "Start working on driver changes", leading to "DriverImplementation?" Can you start the same work twice? :-) It also indicates a route from fast path to driver implementation, which shouldn't be there, since the fast path means no driver changes? Also in one note the fast path is described as "no impact to driver" but in another it's "minor changes" which are two different criteria?

contribute/process/images/abiprocess.pu
```diff
@@ -0,0 +1,25 @@
...
+
+note right of DriverImplementation: If 1) FW change is MAJOR or\n2) MINOR but modifies\nupstream topologies, then\nkernel change must\nbe merged at the same time
```
lyakh, line 19: what does "merged at the same time" mean and how does it help? You can click the two "merge" buttons within the same millisecond but it won't prevent users or distributions from mixing different versions?
"Upstream topologies" are topologies for platforms / configurations available in the kernel mainline (as opposed to only supported in sof-dev) or only those, whose names are used in one of machine drivers? E.g. is only sof-apl-pcm512x.tplg an upstream topology or all other sof-apl-*-pcm512x.tplg too?

contribute/process/images/abiprocess.pu
kv2019i: @lyakh Good question. I This was on purpose. I wanted to highlight that if you go via the "normal path", we can sooner figure out whether a driver change is needed to merge the ABI change. Currenly we have many cases where this realization only happens at the ABI board meeting, which is very late. If you take the "fast path", you have the risk that you will be sent back to the drawing board. Makes sense?

contribute/process/images/abiprocess.pu
kv2019i: @lyakh I think we need to keep this a bit vague as we don't have infra to do cross-repo merges. This is not strictly related to the ABI, but rather our CI. Typically kernel side would have to be merged first, or otherwise CI will get broken (new topology files deployed to CI that the kernel does not know how to parse). As it's related to our CI, this consideration applies to all topologies we test in our CI.

kv2019i:
v2-to-3 changes:
- return to a simpler model where driver PR development states are not reflected in the diagram
- remove note on CI considerations (topology with new ABI features deployed to CI will)) -- too confusing to have this in the same diagram
- add more cross-links to ABI version defs

I'll add a picture here later.

kv2019i:
@lgirdwood Maybe we add a "RFC" column to the ABI change board to gather all PRs current in RFC or RFCApproved stages.

cujomalainey:
I like workflow, I am just worried with the speed at which team that process would be followed simply because not everyone is aware. I wonder if something like https://github.com/palantir/policy-bot might come in handy.

lgirdwood:
@kv2019i I would also show where the FW is merged alongside the driver merge.

kv2019i:
@cujomalainey wrote:
> I like workflow, I am just worried with the speed at which team that process would be followed simply because not everyone is aware. I wonder if something like https://github.com/palantir/policy-bot might come in handy.

Thanks for the pointers! Palantir seems very interesting indeed. FYI for @marc-hb and @xiulipan . Based on review of ABI changes during last half year, the biggest slowdowns in this time period were due to unclarity on the criteria. So let's see how much we can smoothen the process by just clarifying the process (and i.e. having a diagram for the process).

lgirdwood:
@kv2019i I think we should make it really obvious that the FW + driver PR owner(s) are wholly responsible for "driving" the process. i.e. it's their responsibility to bring in the correct people at each stage and get approvals from kernel/FW folks. i.e. I would want to see kernel approval on the FW PR and vice versa, this will save maintainers having to check with each other.

kv2019i:
> @kv2019i I think we should make it really obvious that the FW + driver PR owner(s) are wholly 
> responsible for "driving" the process. i.e. it's their responsibility to bring in the correct people at > each stage and get approvals from kernel/FW folks. i.e. I would want to see kernel approval on 
> the FW PR and vice versa, this will save maintainers having to check with each other.

Yes, especially with both driver and FW maintainers (of which there are many) involved in the process, it doesn't scale if the maintainers need to chase after reviews.

I now added text to this effect to the section outlining the process. Let me know if you have ideas how to emphasize this even more.

kv2019i:
@cujomalainey 
> Maybe part of the PR process would be beneficial to explicitly name the people who are handling ABI review (i.e. add it to the template)

Ack. I now added separate sections that point to the list of TSC members and list of driver maintainers. 

Interesting point though. I found from TSC minutes that ABI changes should be approved by TSC, but hmm, is that really so? I now put wording such that "approval from a TSC member is required".

kv2019i:
> @kv2019i I would also show where the FW is merged alongside the driver merge.

The process diagram is getting somewhat unwieldy already. How about this (picture generated from the commit I pushed today):

plantuml-de09125c5fb2b7d7264b3ca52fd7c9ef0b231829

contribute/process/images/abiprocess.pu
cujomalainey: In what context is a change to the driver no required? If the ABI is being updated won't it need it in both places? Or is this referring to a backwards compatible change. (e.g. adding to an enum)

lgirdwood:
@kv2019i one more thing, I've now created an ABI milestones to work alongside the classifier. This makes the ABI classifier output visible on the PR and shows any ABI change history. i.e. I've  had to move a lot from 3.16 -> 3.17 today. You should see this on kernel PRs now too.

contribute/process/images/abiprocess.pu
kv2019i: @cujomalainey wrote:
> In what context is a change to the driver no required? If the ABI is being updated won't it 
> need it in both places? Or is this referring to a backwards compatible change. (e.g. adding 

Indeed these would be MINOR changes, which are backwards compatible. If we bump MAJOR version, then driver change is mandatory.

What has been muddying the picture are MINOR ABI changes, but where part of the FW change takes new features into use in the SOF topologies used in CI. This was now taken out from the ABI process description as it's not really ABI management anymore, but a question of how the CI(s) is/are managed. In many cases driver change is still optional, but e.g. if you now add a new component type (pre-UUID system), kernel must be modified to understand the new component. With the UUID based component creation in place, we will have less and less cases like this.

lgirdwood:
> Looks good, two minor questions, even if they are no then I still approve
> 
> 1. do we want to add in the UML that PRs for firmware and driver must land at the same time? Just to keep ABIs relatively in sync across trees.

I think so, we dont want too much time between kernel and FW merge of PRs.

> 2. Given the ABI minor tracker is already quite full, how would we go about flushing it? Would we remove items from the tracker once they are part of a release?

I leave them in the tracker for history. I manually have to move things around but we could probably improve this my having a column for merged items.

kv2019i:
@lgirdwood wrote:
> @cujomalainey wrote:
>>     1. do we want to add in the UML that PRs for firmware and driver must land at the same time? Just to keep ABIs relatively in sync across trees.
> I think so, we dont want too much time between kernel and FW merge of PRs.

Hmm. How about we simply enforce a rule that _if_ driver PR is required (not mandatory in the new process for MINOR update), then driver PR must be merged first. 

Driver side will have capability to support multiple MAJOR versions. And if we some CI-reason to update the kernel for a MINOR update, it makes sense to merge the driver first as well.

If ok, I can update this to the diagram?

cujomalainey:
> @lgirdwood wrote:
> 
> > @cujomalainey wrote:
> > > ```
> > > 1. do we want to add in the UML that PRs for firmware and driver must land at the same time? Just to keep ABIs relatively in sync across trees.
> > > ```
> > 
> > 
> > I think so, we dont want too much time between kernel and FW merge of PRs.
> 
> Hmm. How about we simply enforce a rule that _if_ driver PR is required (not mandatory in the new process for MINOR update), then driver PR must be merged first.
> 
> Driver side will have capability to support multiple MAJOR versions. And if we some CI-reason to update the kernel for a MINOR update, it makes sense to merge the driver first as well.
> 
> If ok, I can update this to the diagram?

Sounds reasonable to me

deb-intel:
@lgirdwood @cujomalainey Is this PR ready to be merged? Thanks.

cujomalainey:
I think we are waiting for @lgirdwood to respond to the possible change in https://github.com/thesofproject/sof-docs/pull/244#issuecomment-644317032

kv2019i:
@cujomalainey wrote:
> I think we are waiting for @lgirdwood to respond to the possible change in [#244 (comment)](https://github.com/thesofproject/sof-docs/pull/244#issuecomment-644317032)

I'll make an update later today with the proposed additional wording, so @lgirdwood can then go ahead and approve if it looks good.

kv2019i:
Added explicit sequence for ordering of driver and FW change merges. Updated diagram now:

plantuml-00f615f92abd8f6cb978fd574687853c34725384

lgirdwood:
@cujomalainey @kv2019i @deb-intel all good by me. Please merge.

kv2019i:
Thanks all! @deb-intel @lgirdwood who does the publish step (I see old version still online)?

deb-intel:
@kv2019i I am currently making a few grammatical edits to some of the recently merged PRs. Once I merge those PRs, I will publish the SOF site. I am aiming for the end of today (U.S). Do you need to see it published sooner? If so, I can try to accommodate that.

kv2019i:
@deb-intel wrote:
> @kv2019i I am currently making a few grammatical edits to some of the recently merged PRs. Once I merge those PRs, I will publish the SOF site. I am aiming for the end of today (U.S). Do you need to see it published sooner? If so, I can try to

Thanks a lot, no rush with my change! I just wanted to check whether I need to take more actions myself to get the change published.

PR #245, opened 2020-06-01
UUID usage on SOF

keyonjie:
Add a page to describe how to use UUID on SOF, the UUID will be the replacement of the traditional component type.

### DIFF
--- a/developer_guides/firmware/components/component-overview.rst
+++ b/developer_guides/firmware/components/component-overview.rst
@@ -24,10 +24,12 @@ uAPI to create a component of that ``type``. It also provides an entry point
 to the component ops implementation.
 
 UUIDs (Universally Unique Identifiers) provide a more scalable and
-collision-free way of component identification. UUIDs are currently used by
-the tracing subsystem while both the topology and driver still depend on
-``type`` to define topology and create component instances. UUIDs are
-expected to replace the ``type`` in these parts as well.
+collision-free way of component identification. UUIDs are currently used as
+the standard interface by all users of the SOF firmware, including the
+tracing subsystem, the topology .m4 files, and the Linux topology driver.
+Using the ``type`` to define topology and create component is still supported
+today, but the ``type`` could be moved out of the IPC struct in the future,
+so allocating UUID for the new added component driver is *mandatory* now.
 
 The UUID entry declared in the FW code contains the identifier value as well
 as the object which is the component name in this case. Both are
--- a/developer_guides/uuid/index.rst
+++ b/developer_guides/uuid/index.rst
@@ -1,0 +1,129 @@
+.. _uuid:
+
+UUID Usage in SOF
+#################
+
+Why UUID Needed
+***************
+
+To develop a new audio signal processing component, we traditionally
+implemented the component driver with a new unique component type
+introduced. The pain points with this method include how to keep the
+ABI (Application Binary Interface) aligned for the topology file, the
+driver, and the firmware. And if unaligned, how should we make them
+backward compatible. For example, When adding a new component, we could
+have 8 types of combination of topology/driver/firmware versions with
+either the new component is supported or not in each
+topology/driver/firmware part. Even more, if enumerated type is used
+for the component type and the sequence in the enumerate list
+(the value) is changed during an update, there could be component type
+collision if versions are not aligned.
+
+UUIDs (Universally Unique Identifiers) provide a more scalable and
+collision-free way of component identification. UUIDs are used as the
+standard interface by all users of the SOF firmware, including the
+tracing subsystem, the topology .m4 files, and the Linux topology
+driver.
+
+Allocating a new UUID for the new added component is recommended, as
+the component type will be replaced by UUID in the IPC structures,
+in the future. For the whole SOF stack, the usage of the UUID shall
+follow rules as below:
+
+UUID Allocation
+***************
+The UUID allocation of a specific component shall use the version 4 in
+`UUID wikipedia <https://en.wikipedia.org/wiki/Universally_unique_identifier>`__,
+and the value shall be declared in the firmware component driver with
+``DECLARE_SOF_RT_UUID``, for details, please refer to the API
+documentation :ref:`uuid-api`.
+
+UUID in Topology
+****************
+The same UUID shall be used in topology .m4 file for a new added
+widget (corresponding to the new added component), since we have
+implemented macro to help to handle the conversion task, just use the
+exactly same macro ``DECLARE_SOF_RT_UUID`` as depicted in the firmware
+source, e.g for SRC component the below shall be added to the topology
+.m4 tools/topology/m4/src.m4:
+
+.. code-block:: none
+
+   DECLARE_SOF_RT_UUID("host", host_uuid, 0x8b9d100c, 0x6d78, 0x418f,
+        0x90, 0xa3, 0xe0, 0xe8, 0x05, 0xd0, 0x85, 0x2b);
+
+Linux Topology Driver
+*********************
+The topology driver will parse the 16-byte UUID token, append it to the
+extended data of the IPC struct, and sent it to the
+firmware in component creation stage, **for all components**.
+
+.. code-block:: none
+
+   /* Component extended tokens */
+   static const struct sof_topology_token comp_ext_tokens[] = {
+    {SOF_TKN_COMP_UUID,
+        SND_SOC_TPLG_TUPLE_TYPE_UUID, get_token_uuid,
+        offsetof(struct sof_ipc_comp_new_ext, uuid), 0},
+   };
+
+.. code-block:: none
+
+   static int sof_widget_ready(struct snd_soc_component *scomp, int index,
+                               struct snd_soc_dapm_widget *w,
+                               struct snd_soc_tplg_dapm_widget *tw)
+   {
+   ...
+           ret = sof_parse_tokens(scomp, &swidget->comp_ext, comp_ext_tokens,
+                                  ARRAY_SIZE(comp_ext_tokens), tw->priv.array,
+                                  le32_to_cpu(tw->priv.size));
+   ...
+   }
+
+UUID Arrays Stored Section
+**************************
+Only the UUID arrays for component types used in topology file are
+stored to the .rodata section as static data, for limited memory
+footprint purpose, e.g.
+19 component types * 16 Bytes/component type = 304 Bytes.
+
+UUID to Component Driver Mapping
+********************************
+The firmware will use UUID byte array to match the component driver, if
+it is provided from the Linux driver side, otherwise, fallback to use
+the traditional component type for backwards-compatible behavior.
+
+.. code-block:: none
+
+   static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
+   {
+        ...
+        /* validate the extended data */
+        ...
+        /* use UUID to match the driver if UUID is provided */
+        if (comp->ext_data_offset) {
+            /* use component type if old tplg without UUID used */
+            if (sof_is_uuid_nil(comp_ext->uuid))
+                goto comp_type_match;
+
+            /* search driver list with UUID */
+            ...
+            /* matched, return drv */
+            return drv;
+
+            /* not found, failed */
+            return NULL;
+        }
+
+   comp_type_match:
+        /* search driver list for driver type */
+        ...
+        /* return the component type matched driver */
+        return drv;
+   }
+
+ABI Alignment for UUID Support
+******************************
+In general, UUID will be used only all FW/tplg/driver are in ABI version
+equal or greater than 3.17, otherwise component type will be used.
+
--- a/developer_guides/index.rst
+++ b/developer_guides/index.rst
@@ -13,6 +13,7 @@ terminology before reading further.
    firmware/index
    unit_tests
    topology/topology
+   uuid/index.rst
    debugability/index
    tuning/sof-ctl
    rimage/index.rst


developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,118 @@
...
+   struct snd_soc_tplg_vendor_uuid_elem {
+           __le32 token;
+           char uuid[16];
+   } __attribute__((packed));
+
+Linux topology driver
+*********************
+The topology driver will parse the 16Bytes UUID token and sent it to the
+firmware in component creation stage, **for all components**.
+
```
lgirdwood, line 49: Why does the driver care about the UUID ? It should be passing this direct to the FW. The driver only cares about the type atm for IPC1.  In IPC2.0 the driver will only care about type (but with less enums).

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,118 @@
...
+1~4KB (depends on how many modules are compiled) memory consumption
+increase. e.g. in src/platform/cannonlake/cannonlake.x.in:
+
+.. code-block:: none
+
+  .fw_ready : ALIGN(4)
+  {
+    KEEP (*(.fw_ready))
+    KEEP (*(.fw_ready_metadata))
+  } >sof_fw :sof_fw_phdr
```
ktrzcinx, line 87: False, UUID are going to be stored in:
```
  .static_uuid_entries : ALIGN(4)
  {
  *(*.static_uuids)
  } >sof_data :sof_data_phdr
```

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,118 @@
+.. _uuid:
+
+UUID Usage in SOF
+#################
+
+We use universally unique identifier (UUID) for component type
+definition in SOF, and for the whole SOF stack, the UUID usage is
+following below rules:
```
kv2019i, line 8: Thanks @keyonjie for the patch. One generic comment that bothers me a bit is that UUID is really part of the solution. From a developer point view, developer doesn't really have a problem that "I want to use UUID". Rather developers want to add components and use SOF infra to describe their component's interface. UUID is one aspect SOF provides to solve this. I think it's ok to have a separate section that summarizes the UUID implementation and semantics, but looking at the painpoints of our recent discussions w.r.t. component type and UUID, there are questions that go outside the UUID discussion.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,118 @@
...
+                                  ARRAY_SIZE(comp_tokens), tw->priv.array,
+                                  le32_to_cpu(tw->priv.size));
+   ...
+   }
+
+UUID arrays stored section
+**************************
+The UUID arrays are stored to SRAM in firmware, though it will lead to a
+1~4KB (depends on how many modules are compiled) memory consumption
+increase. e.g. in src/platform/cannonlake/cannonlake.x.in:
```
plbossart, line 79: do we actually need to store the UUID in the firmware?
Wondering if we could optimize this just like we do for strings with traces, the firmware only contains an index and the strings are extracted by indirection. We could do something similar on firmware generation where a specific UUID is replaced by a 'dynamic index', with e.g. the mapping between UUID and index in the firmware manfest. It's kind of clunky if the fimware needs to do tests with 10 bytes to figure out what to do with the IPC.

developer_guides/uuid/index.rst
keyonjie: @lgirdwood are you meaning that not using the generic UUID token, but to create another vendor specific set of tokens for SOF struct sof_uuid (we might need 1 word, 2 short, and 1 string token) ? The driver here is not caring the UUID, it only parsing it and construct the IPC new structure which will be sent for component creation.

plbossart: no, we have to use the full-blown UUID to avoid collisions between vendors.

developer_guides/uuid/index.rst
keyonjie: oops, copy-past error, it should be:
```
+  .static_uuid_entries : ALIGN(4)
+  {
+  *(*.static_uuids)
+  } >sof_fw :sof_fw_phdr
```

Thanks for spotting this.

developer_guides/uuid/index.rst
keyonjie: This was actually my question also, That will be the 2nd solution in my previous comment here: https://github.com/thesofproject/sof/pull/2956#issuecomment-631833396

I think I was enlightened during the discussion that this solution might hit issues during dynamic module/component load/unload, the pre-built FW extended manifest might be inconsistent or unaligned between the UUID <-->index pairs. 

@lgirdwood @kv2019i correct me if I am wrong.

keyonjie:
> I think this is good start, but I think we need to level-up a bit and tackle the topic from a component developers point of view. This is now partly in conflict with:
> firmware/components/component-overview.rst

Yes, there should be joint update in the documentation of the component part.

> 
> ... we'd need to write down how component_type and UUID relate to each other. I think that's where most of the arguments stem from (and reason to introduce UUIDs in general).

Let me add a chapter to talk about this. Thanks for the points.

developer_guides/uuid/index.rst
keyonjie: > no, we have to use the full-blown UUID to avoid collisions between vendors.

OK, then paring the 16B UUID token here looks good to me @lgirdwood ?

developer_guides/uuid/index.rst
plbossart: We're going to hit issues with module load/unload as well for string handling for traces...

kv2019i: @plbossart The specific idea you raise matches with the initial design of a UUID dictionary, and it was also implemented, but rejected for SOF, see @lgirdwood 's guidance in https://github.com/thesofproject/sof/pull/2956#issuecomment-631978166 
We have to pick a path one way or the other...

lgirdwood: @keyonjie where does 1 - 4KB come from ? UUID means we are adding 16 bytes per component.
```grep -rn DECLARE_SOF_UUID --exclude *.lst --exclude *.h | wc -l
35
```
This also includes some duplicates from different platforms so this number is nearer 30 and about 500 bytes. If we are near 1kB or more then it means we rea duplicating UUIDs somewhere.

lgirdwood: I've found the reason why UUIDs are growing - 69 on master now. They should only be used for components that are loaded via topology, but they seem to being used for kernel subsystems now.  @mmaka1 why are we using UUIDs on kernel subsystems now ? I would expect kernel to
```c
/* system UUIDs are stored in LDC */
DECLARE_SYSTEM_UUID("sys.mem.alloc", "0x1234...........");
```
If userspace needs UUIDs for all kernel objects then it can read them from LDC.

keyonjie: Yes, and remember we have name(max 32 Bytes more) stored for each UUID also, we actually store 48 Bytes for each entry, so that is 48 * 69 = 3312B > 3KB.
```
struct sof_uuid_entry {
	struct sof_uuid id;
	const char name[UUID_NAME_MAX_LEN];
};
```

keyonjie: > We're going to hit issues with module load/unload as well for string handling for traces...

That's true, I think @mmaka1 is considering changing that for trace also.

plbossart: > Yes, and remember we have name(max 32 Bytes more) stored for each UUID also, we actually store 48 Bytes for each entry, so that is 48 * 69 = 3312B > 3KB.
> 
> ```
> struct sof_uuid_entry {
> 	struct sof_uuid id;
> 	const char name[UUID_NAME_MAX_LEN];
> };
> ```

why would you need to pass the name to the firmware? is it necessary?

keyonjie: > > Yes, and remember we have name(max 32 Bytes more) stored for each UUID also, we actually store 48 Bytes for each entry, so that is 48 * 69 = 3312B > 3KB.
> > ```
> > struct sof_uuid_entry {
> > 	struct sof_uuid id;
> > 	const char name[UUID_NAME_MAX_LEN];
> > };
> > ```
> 
> why would you need to pass the name to the firmware? is it necessary?

No, I am here only stating the fact that the whole 48 Bytes are stored in and occupying the SRAM space, see it here:
```
struct sof_uuid_entry {
	struct sof_uuid id;
	const char name[UUID_NAME_MAX_LEN];
};
...
#define DECLARE_SOF_UUID(entity_name, uuid_name,			\
			 va, vb, vc,					\
			 vd0, vd1, vd2, vd3, vd4, vd5, vd6, vd7)	\
	__section(".static_uuids")					\
	static const struct sof_uuid_entry uuid_name = {		\
		{.a = va, .b = vb, .c = vc,				\
		 .d = {vd0, vd1, vd2, vd3, vd4, vd5, vd6, vd7}},	\
		entity_name "\0"					\
	}
```

We are only passing the 16Bytes (corresponding to the 'id' in struct sof_uuid_entry) from driver to FW.

keyonjie:
> > I think this is good start, but I think we need to level-up a bit and tackle the topic from a component developers point of view. This is now partly in conflict with:
> > firmware/components/component-overview.rst
> 
> Yes, there should be joint update in the documentation of the component part.
> 
> > ... we'd need to write down how component_type and UUID relate to each other. I think that's where most of the arguments stem from (and reason to introduce UUIDs in general).
> 
> Let me add a chapter to talk about this. Thanks for the points.

updated now. @kv2019i

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Allocating a new UUID for the new added component is mandatory, and
+UUIDs are expected to replace the component type in the near future. For
+the whole SOF stack, the usage of UUID should follow rules as below:
+
+UUID allocation
+***************
+The UUID allocation of a specific component should use the version 4 in
+https://en.wikipedia.org/wiki/Universally_unique_identifier, and the
+value should be specified in the firmware component dirver, e.g. for src
```
mmaka1, line 29: This second part might be replaced by a reference to the tutorial to maintain a single copy of code samples (easier to maintain).

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+   ...
+           ret = sof_parse_tokens(scomp, &template, comp_tokens,
+                                  ARRAY_SIZE(comp_tokens), tw->priv.array,
+                                  le32_to_cpu(tw->priv.size));
+   ...
+   }
+
+UUID arrays stored section
+**************************
+The UUID arrays are stored to SRAM as a separated section, though it
```
mmaka1, line 94: A general documentation should avoid references to specific memory types since they may be completely different for various platforms. And the whole concept of the dedicated section is that it is outside the run-time memory space even (SRAM on some systems).

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+consumption increase. e.g. in src/platform/cannonlake/cannonlake.x.in:
+
+.. code-block:: none
+
+  .static_uuid_entries : ALIGN(4)
+  {
+    *(*.static_uuids)
+  } >sof_fw :sof_fw_phdr
+
+UUID to component driver mapping
```
mmaka1, line 105: This section should be replaced with a reference to the tutorial to avoid duplicates since a complete code example is available and maintained there.

developer_guides/uuid/index.rst
kv2019i: Updating offline discussions that:
 - name[UUID_NAME_LAX_LEN] will not be stored
 - only audio components visible in topology will have their uuids stored in FW memory
These both will save the memory footprint.

keyonjie: > Updating offline discussions that:
> 
> * name[UUID_NAME_LAX_LEN] will not be stored
> * only audio components visible in topology will have their uuids stored in FW memory
>   These both will save the memory footprint.

Thanks @kv2019i let me update the doc accordingly.

developer_guides/firmware/components/component-overview.rst
```diff
@@ -24,10 +24,14 @@ uAPI to create a component of that ``type``. It also provides an entry point
...
 UUIDs (Universally Unique Identifiers) provide a more scalable and
-collision-free way of component identification. UUIDs are currently used by
-the tracing subsystem while both the topology and driver still depend on
-``type`` to define topology and create component instances. UUIDs are
-expected to replace the ``type`` in these parts as well.
+collision-free way of component identification. UUIDs are currently used as
+the standard interface by all users of the SOF firmware, including the
+tracing subsystem, the topology .m4 files, and the Linux topology driver.
+We still support using the ``type`` to define topology and create component
+instances at the moment, but UUIDs are expected to replace the ``type`` in
```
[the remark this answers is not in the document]
keyonjie: It's true, the type will be replaced. per my understanding.

developer_guides/firmware/components/component-overview.rst
```diff
@@ -24,10 +24,14 @@ uAPI to create a component of that ``type``. It also provides an entry point
...
+the near future, so UUID allocation for the new added component driver is
+mandatory. For more details about how to use UUID with SOF, see
+https://thesofproject.github.io/latest/developer_guides/uuid/index.html
```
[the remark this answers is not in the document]
keyonjie: um, here I actually want to provide a link to help to understand how the UUID works with the full SOF stack (among tplg/driver/fw).  How about wording it like this:
For more details about how the UUID works in SOF, please refer to...

developer_guides/uuid/index.rst
keyonjie: Thanks. Since we have decided to split and save only part of those UUIDs to rodata section, I would need to refine this part anyway.

developer_guides/uuid/index.rst
keyonjie: Yes, let me move the content to the tutorial and provide a reference here only.

developer_guides/firmware/components/component-overview.rst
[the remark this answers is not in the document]
plbossart: it's not my understanding. You need a process/decoder/dai/etc type so that the firmware has a high-level understanding of what the component is.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+
+UUID Usage in SOF
+#################
+
+Why UUID is needed
+******************
+
+When developing a new audio signal processing component, we
+traditionally implement the component driver with a new unique component
+type declared together, the painpoints with this method including ABI
```
plbossart, line 11: not sure what you meant by 'declared together'. together with what?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+alignment among topology/driver/firmware for each new added components,
+and there could be collision if versions are not aligned.
```
plbossart, line 13: no, it's a collision between different vendors using the same type for different usages and component.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+
+UUIDs (Universally Unique Identifiers) provide a more scalable and
+collisionn-free way of component identification. UUIDs are currently
```
plbossart, line 16: typo: collision

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+used as the standard interface by all users of the SOF firmware,
+including the tracing subsystem, the topology .m4 files, and the Linux
+topology driver.
+
+Allocating a new UUID for the new added component is mandatory, and
+UUIDs are expected to replace the component type in the near future. For
```
plbossart, line 22: no. we need to keep generic types.

developer_guides/uuid/index.rst
plbossart, line 29: typo: driver

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+****************
+We should provide UUID string in topology .m4 file in the exactly same
+format as commented in the firmware, e.g for host it is
+c1c5326d-8390-46b4-aa47-95c3beca6550, so in the topology .m4 file
+tools/topology/m4/src.m4:
+
+.. code-block:: none
+
+   dnl Defines the macro for SRC widget
+   define(SRC_UUID_ARRAY, UUID_ARRAY("c1c5326d-8390-46b4-aa47-95c3beca6550"))
```
plbossart, line 48: this makes no sense to me, why not use the same format as line 35 and 36 so that we can copy/paste between firmware and topology?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+.. code-block:: none
+
+   /* Generic component tokens */
+   static const struct sof_topology_token comp_tokens[] = {
+           {SOF_TKN_COMP_CORE_ID,
+               SND_SOC_TPLG_TUPLE_TYPE_WORD, get_token_u32,
+               offsetof(struct sof_ipc_comp, core), 0},
+           {SOF_TKN_COMP_UUID,
+               SND_SOC_TPLG_TUPLE_TYPE_UUID, get_token_uuid,
+               offsetof(struct sof_ipc_comp, uuid), 0},
```
plbossart, line 76: can you have a component with the same uuid on different cores then?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+   };
+
+.. code-block:: none
+
+   static int sof_widget_ready(struct snd_soc_component *scomp, int index,
+                               struct snd_soc_dapm_widget *w,
+                               struct snd_soc_tplg_dapm_widget *tw)
+   {
+   ...
+           ret = sof_parse_tokens(scomp, &template, comp_tokens,
```
plbossart, line 86: what is template?

developer_guides/uuid/index.rst
plbossart, line 94: typo: separate

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,134 @@
...
+  .static_uuid_entries : ALIGN(4)
+  {
+    *(*.static_uuids)
+  } >sof_fw :sof_fw_phdr
+
+UUID to component driver mapping
+********************************
+The firmware will use UUID byte array as the first choice to match the
+component driver, and fallback to use traditional component type if
+failed:
```
plbossart, line 109: why? we don't need to support the old kernel/new firmware model. it'd be perfectly fine to *require* uuids to be provided for each component.

developer_guides/uuid/index.rst
kv2019i: @plbossart I think this is similar use of "template" as in soc-dapm.c -- a base class of sorts.

developer_guides/uuid/index.rst
kv2019i: @plbossart Do we see value in being able to define topologies with component type. E.g. one could define a topology with a volume component. Without uuid, it would pick whatever implementation the DSP has (component_type==VOLUME). These topologies would work with multiple imlementations of volume.

plbossart: @kv2019i that sounds like asking for trouble? If you have multiple implementations of the volume, then you may have different parameters or configurations, so likely to have the driver provide a config that the firmware doesn't understand. I would really err on the side of strong-types with no ambiguity: topology, firmware and driver all have a common understanding of what each component on each path do and how to configure them. I think it's even more important for 3rd party, e.g. you may have a plugin version on one path and a different one on the other due to MCPS/memory/performance/acoustics requirements.

lgirdwood: "type" will be deprecated and removed with IPC2.0, this just give some overlap.
@keyonjie type will only be used if UUID is missing, if UUID fails then there is no match.

plbossart: @lgirdwood 'type' as a component identifier will no longer be used, but I think we need to keep a 'type' or 'class' that tells us what a component is (DAI, effect, decoder, etc). Maybe you don't pass it over IPC but it's part of the component definition.

plbossart: And second @lgirdwood why do we need a fallback mechanism? we can bump the firmware and expect the kernel to use the UUIDs.

lgirdwood: @plbossart agree, driver still needs some limited types for DAPM. The falback was to help with combinations of different ABIs between FW, topology and driver. Removing type from IPC ABI would be a MAJOR and it would be best to do this with IPC2.0 MAJOR.

developer_guides/uuid/index.rst
keyonjie: Looks I am making semantic wrong here, I meant that we need to define new component type in the firmware (and shared with the driver) together with the component driver implementation itself.

developer_guides/uuid/index.rst
keyonjie: Thanks.

developer_guides/uuid/index.rst
keyonjie: What I wrote in the commit is only about the collision (or I should call it 'unaligned' actually) of different component type, e.g. in some version the type 20 could mean SMART_AMP but in most versions 20 should be DCBLOCK, this doesn't happen frequently, but if the ABI management is not run good enough (that is why @kv2019i has been proposing improvement on it).

For the collision of different vendors using the same type, we talked about this in the discussion meeting also, we have a thought to introduce UUID array for the component driver, which means to list all supported vendors' implementation to the driver (something like the supported PCI IDs in our linux PCI device driver) and use that for the component driver matching. With this, we can still use a single UUID for a specific component type, and no class/subtype mechanism needed. This is still in discussion stage yet, as we don't have real different vendors's implementation for the same driver upstream yet, but we do think of it and leave the possibility of adding support to it.

developer_guides/uuid/index.rst
keyonjie: @kv2019i @plbossart I don't see the mandatory of keeping the component type, e.g. for volume component, it is just like another component which we created a kcontrol for it in the driver side, we use the component id (actually an unique instant id among all kinds of components) for setting/getting the values b/w firmware and driver.

Can you help to share an example which type we have to keep?

developer_guides/uuid/index.rst
keyonjie: Eagle eyes! I just read it 5 more times before I can find this typo.

developer_guides/uuid/index.rst
keyonjie: I think yes, but that is another topic no?

developer_guides/uuid/index.rst
keyonjie: Yes "template" is widely use naming in the asoc core driver (e.g. soc-dapm.c, soc-topology.c), maybe we should keep it the function locally (and not in the function argument), that's what I have changed in the kernel PR.

developer_guides/uuid/index.rst
keyonjie: Thanks.

developer_guides/uuid/index.rst
keyonjie: Yes, this is only try to give some overlap during the transition period, somebody might use old tplg files, or UUID of some component may not be specified in the tplg files, we will throw out a warning for this kind of mistake at the moment, and we can remove this fallback mechanism when we should.

developer_guides/firmware/components/component-overview.rst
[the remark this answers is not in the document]
keyonjie: > it's not my understanding. You need a process/decoder/dai/etc type so that the firmware has a high-level understanding of what the component is.

I get your point, if the Linux driver don't need to understand the component type, we can move it out from the struct sof_ipc_comp (e.g. to struct comp_dev), this will allow checking the component type inside firmware such as what comp_get_endpoint_type() do.

developer_guides/uuid/index.rst
kv2019i: @keyonjie For example, without component type, how would the kernel driver know that a it should parse "asrc_tokens" when creating the ASRC component, and use "sof_ipc_comp_asrc" struct to create the IPC message? Similarly, how to know when to parse "volume_tokens". For components that have only a blob as a parameter, this is not needed and we can use the generic "effect" component_type. E.g. we could have more robust check e.g. in snd_sof_volume_put() to ensure volume IPC is sent to an actual volume component -- now we rely on the topology to be correct here. If we start exposing more parameters via kcontrols, I'm not sure this scales if kernel is just passing blobs. Is your thinking to move all these to "effects"?

developer_guides/uuid/index.rst
kv2019i: @lgirdwood @plbossart Do we agree that if we change semantics of "component_type", this is a MAJOR bump? I know we haven't committed to support new fw with old kernels, but I think we should stick to semantic versioning and I think this is a MAJOR bump. OTOH, +95% of users will update the topologies at the same time, so in practise SOF release will require a very recent kernel to use.

developer_guides/uuid/index.rst
keyonjie: @kv2019i we use snd_soc_dapm_type (specified in tplg for each dapm widget) to decide if it is asrc, pga, or effect, we don't use the component type there.

developer_guides/uuid/index.rst
lgirdwood: if UUID is not present then "type" is used using existing IPC
if UUID is present in topology then we use new IPC message (MINOR update)
if UUID is present in topology and FW is too old we use "type" and old message (MINOR).

developer_guides/uuid/index.rst
keyonjie: line 35 and 36 is C macro to define struct sof_uuid_entry, but ASoC UUID token is u8 array[16], what did you suggest to copy/paste? the "0xc1c5326d, 0x8390, 0x46b4, 0xaa, 0x47, 0x95, 0xc3, 0xbe, 0xca, 0x65, 0x50)"? The UUID standard (see https://en.wikipedia.org/wiki/Universally_unique_identifier) string "c1c5326d-8390-46b4-aa47-95c3beca6550" (copy/paste from the line 34) looks more compact and reasonable, no?

plbossart: I don't care about the format, I just want the firmware files and topoligy files to use the same format to avoid having to edit bytes.

developer_guides/uuid/index.rst
kv2019i: @keyonjie Ok, I get your point. But snd_soc_dapm_type is effectively the component type then. I think this should be reflected in rest of the docs. E.g. developer_guides/firmware/component-tutorial/tut-ii-topology.rst does not touch on what is the meaning of the dapm_type, but only covers component_type. After this change, there is a 1:1 mapping between the two and component author needs to pick the dapm-type for the new component.

kv2019i: @plbossart Ok with above? I think dapm/comp_type is more of a detail, I'm happy as long as we keep a generic "class" notion in kernel (and not have any per-UUID action in kernel).

developer_guides/uuid/index.rst
keyonjie: > I don't care about the format, I just want the firmware files and topology files to use the same format to avoid having to edit bytes.

That was what @lgirdwood commented, we already support copying from the firmware files and pasting to the topology .m4 files directly, no need to edit bytes.

developer_guides/uuid/index.rst
keyonjie: We have 8 combination cases here so it is actually somewhat complicate, if we can use major bump and Force "driver ABI major == FW ABI major", it will make our life easier.


FW | tplg | driver | supported today? | Force driver ABI >= FW ABI | Use major bump && Force driver ABI major == FW ABI major | comment
-- | -- | -- | -- | -- | -- | --
old | old | old | Y | support | support |  
old | old | new | support if component type fallback support in driver with old ABI | support if component type fallback support in driver with old ABI | rejected |  
old | new | old | support, uuid in tplg doesn’t take effect | support, uuid in tplg doesn’t take effect | support, uuid in tplg doesn’t take effect |  
old | new | new | support if component type fallback support in driver with old ABI | support if component type fallback support in driver with old ABI | rejected |  
new | old | old | N, buggy. | rejected | rejected |  
new | old | new | support, need support component type fallback in both fw&driver | support, need support component type fallback in both fw&driver | support, need support component type fallback in both fw&driver |  
new | new | old | N, buggy. | rejected | rejected |  
new | new | new | Y | support | support |

kv2019i: @keyonjie Ok, that's a very thorough table.  I think you are muddying the waters w.r.t. to our agreed semantics for ABI versioning here though:
 - If driver ABI version is forced to be same-or-newer than FW, what's the point to maintain MINOR versions at all?
 - I don't get the "supported today?" column... do you mean the option that UUID is added to the IPC as a MINOR option? 
- if yes to the above, I don't understand the "N, buggy" entries. there's still the comp_type, so I don't see what the bugs would be

I'd prefer to make this a simple call between 
- a)  UUID is a minor bump (we can support old IPC interface in FW), and
- b) UUID is a major bump (we cannot support the old IPC)

a or b, but let's not start changing our versioning semantics mid-flight.

developer_guides/uuid/index.rst
plbossart: Line 35-36 and line 48 show a different format so there's no copy-paste possible between the firmware code and the topology file.

developer_guides/uuid/index.rst
keyonjie: > @keyonjie Ok, I get your point. But snd_soc_dapm_type is effectively the component type then. I think this should be reflected in rest of the docs. E.g. developer_guides/firmware/component-tutorial/tut-ii-topology.rst does not touch on what is the meaning of the dapm_type, but only covers component_type. After this change, there is a 1:1 mapping between the two and component author needs to pick the dapm-type for the new component.

Yes, that's true, user have to specify a dapm type in .m4 when asking for creating a dapm widget.

developer_guides/uuid/index.rst
keyonjie: > Line 35-36 and line 48 show a different format so there's no copy-paste possible between the firmware code and the topology file. the fact that the internal representation is identical with a 16-byte array is irrelevant, what we need to have is the same macro with the same split between bytes.

If you can treat Line *34*(the comment line) and line 48, they are in same format. As you know, here they are C and m4, it is not easy to implement the C structure definition in m4 with the same macros. @plbossart are you suggesting something like this?
```
UUID_ARRAY("0xc1c5326d, 0x8390, 0x46b4, 0xaa, 0x47, 0x95, 0xc3, 0xbe, 0xca, 0x65, 0x50")
```

developer_guides/uuid/index.rst
keyonjie: > @keyonjie Ok, that's a very thorough table. I think you are muddying the waters w.r.t. to our agreed semantics for ABI versioning here though:
> 
> * If driver ABI version is forced to be same-or-newer than FW, what's the point to maintain MINOR versions at all?

I am not sure if we already have assumption that the driver will be newer than the FW in most of the time, from the code, by default if CONFIG_SND_SOC_SOF_STRICT_ABI_CHECKS is not specified, we will only throw a warning only if the FW is newer. This could be a risk, in this UUID case, if FW is newer and try to read the UUID bytes from an old driver IPC struct, out of bounds memory access could happen. From this angle, do you think we should do a major bump for UUID update?
```
	if (v->abi_version > SOF_ABI_VERSION) {
		if (!IS_ENABLED(CONFIG_SND_SOC_SOF_STRICT_ABI_CHECKS)) {
			dev_warn(sdev->dev, "warn: FW ABI is more recent than kernel\n");
		} else {
			dev_err(sdev->dev, "error: FW ABI is more recent than kernel\n");
			return -EINVAL;
		}
	}
```


> * I don't get the "supported today?" column... do you mean the option that UUID is added to the IPC as a MINOR option?

Yes.

> * if yes to the above, I don't understand the "N, buggy" entries. there's still the comp_type, so I don't see what the bugs would be

If we keep the comp_type in the struct sof_ipc_comp and append the uuid to the end, looks we are fine in most of the time (with only MINOR bump), but if the comp_type is not come from the driver anymore (e.g. the comp_type is moved out of the struct sof_ipc_comp) then we have to do a major bump, otherwise, it will be buggy.

> 
> I'd prefer to make this a simple call between
> 
> * a)  UUID is a minor bump (we can support old IPC interface in FW), and
> * b) UUID is a major bump (we cannot support the old IPC)
> 
> a or b, but let's not start changing our versioning semantics mid-flight.

Makes sense, if we are aligned to keep the comp_type in the struct sof_ipc_comp and append the uuid to the end at this moment, we can do only a minor bump with a, and do major bump b maybe in some day that the comp_type is moved out of the struct sof_ipc_comp .

lgirdwood: @keyonjie we will include **both** the type and UUID in topology until the IPC2.0 MAJOR bump. This means we are good for any combination in teh table, the fallback is always to use the type and only use UUID is supported by all FW, driver and topology.

developer_guides/uuid/index.rst
plbossart: This is really simple. We either use

format a) c1c5326d-8390-46b4-aa47-95c3beca6550
format b) 0xc1c5326d, 0x8390, 0x46b4,
                    0xaa, 0x47, 0x95, 0xc3, 0xbe, 0xca, 0x65, 0x50

Pick one for both firmware and M4 macro, don't ask people to format themselves.

lgirdwood: @keyonjie I think we agreed 
1) to use the same definition in FW and topology (to support simple copy and paste).
2) to use the same binary ABI representation between topology and FW.
Please setup a call with me if unsure.

developer_guides/uuid/index.rst
keyonjie: @lgirdwood proposed a solution to fix the ABI backward compatible issue to me, it basically put the UUID array to append to the whole component structure, and use the sof_ipc_comp.reserved member as its offset to the component struct.

keyonjie:
have been busy on some other PM related issues, will update this UUID tomorrow. @lgirdwood @kv2019i

developer_guides/uuid/index.rst
keyonjie: mark this resolved since we are aligned.

developer_guides/uuid/index.rst
keyonjie: mark this resolved since this is addressed.

developer_guides/uuid/index.rst
keyonjie: This is updated, mark it as resolved to make review easier.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/uuid/index.rst
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/firmware/components/component-overview.rst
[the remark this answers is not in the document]
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

developer_guides/firmware/components/component-overview.rst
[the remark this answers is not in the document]
keyonjie: the PR is updated with this part changed, mark it 'resolved'.

keyonjie:
@kv2019i @plbossart @lgirdwood @mmaka1 @ktrzcinx update to reflect to the latest code change in the PRs, please help to review, thanks.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Why UUID Needed
+***************
+
+To develope a new audio signal processing component, we traditionally
+implement the component driver with a new unique component type
+introduced, the painpoints with this method including how to keep ABI
+aligned and backward compatible for different version combination of
+topology/driver/firmware for each new added components, there could be
+component type collision if versions are not aligned.
```
plbossart, line 14: consider a sentence with fewer than 6 lines?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+
+UUIDs (Universally Unique Identifiers) provide a more scalable and
+collision-free way of component identification. UUIDs are currently
```
plbossart, line 17: currently? this is still work in progress?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+used as the standard interface by all users of the SOF firmware,
+including the tracing subsystem, the topology .m4 files, and the Linux
+topology driver.
+
+Allocating a new UUID for the new added component is recommended, as
+the component type might be replaced by UUID in the future. For the
```
plbossart, line 23: so that's unclear now? Is the type replaced yes or no? The previous paragraph strongly hinted that UUID was the only identification method, the use of 'might' makes it a mere possibility. What is the direction?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+whole SOF stack, the usage of UUID should follow rules as below:
+
+UUID Allocation
+***************
+The UUID allocation of a specific component should use the version 4 in
```
plbossart, line 28: shall use. It's a mandatory requirement. should is a recommendation.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+`UUID wikipedia <https://en.wikipedia.org/wiki/Universally_unique_identifier>`__,
+and the value should be delared in the firmware component driver with
```
plbossart, line 30: shall be declared. (fix typo)

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+``DECLARE_SOF_RT_UUID``, for details, please refer to the API
+documentation :ref:`uuid-api`.
+
+UUID in Topology
+****************
+We should provide the same UUID in topology .m4 file for a new added
```
plbossart, line 36: the same UUID shall be used in topology

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+widget (corresponding to the new added component), since we have
+implmented marco to help to handle the converting task, just use the
```
plbossart, line 38: implemented macro - please use an editor with a  spell checker ...

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+exactly same macro ``DECLARE_SOF_RT_UUID`` as depited in the firmware
```
plbossart, line 39: depicted

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+source, e.g for src component the below should be addded to the topology
```
plbossart, line 40: shall

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+                                  ARRAY_SIZE(comp_ext_tokens), tw->priv.array,
+                                  le32_to_cpu(tw->priv.size));
+   ...
+   }
+
+UUID Arrays Stored Section
+**************************
+Only the UUID arrays for runtime created components are stored to the
+.rodata section as static data, it will occupy small memory, e.g.
+19 component types * 16 Bytes/component type = 304 Bytes.
```
plbossart, line 80: this is unclear, this implies there are built-time components so where are those UUIDs stored?

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+           ret = sof_parse_tokens(scomp, &swidget->comp_ext, comp_ext_tokens,
+                                  ARRAY_SIZE(comp_ext_tokens), tw->priv.array,
+                                  le32_to_cpu(tw->priv.size));
+   ...
+   }
+
+UUID Arrays Stored Section
+**************************
+Only the UUID arrays for runtime created components are stored to the
+.rodata section as static data, it will occupy small memory, e.g.
```
plbossart, line 79: , using a limited memory footprint, e.g.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+19 component types * 16 Bytes/component type = 304 Bytes.
+
+UUID to Component Driver Mapping
+********************************
+The firmware will use UUID byte array to match the component driver, if
+it is provided from the Linux driver side, otherwise, fallback to use
+the traditional component type for backward compatible issue.
```
plbossart, line 86: backwards-compatible behavior

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,133 @@
...
+   :widths: 10, 10, 15, 40
+
+   "old", "old", "old", "supported, no UUID"
+   "old", "old", "new", "supported, UUID 0s, FW ignore it"
+   "old", "new", "old", "supported, no UUID"
+   "old", "new", "new", "supported, UUIDs valid, but FW ignore them, component type will be used."
+   "new", "old", "old", "supported, offset/reserved 0 passed to the FW, component type will be used."
+   "new", "old", "new", "supported, UUID 0s, component type will be used in FW."
+   "new", "new", "old", "supported, offset/reserved 0 passed to the FW, component type will be used."
+   "new", "new", "new", "supported, UUID will be supported."
```
plbossart, line 132: can we define old and new here?

keyonjie:
> @keyonjie the technical part seems ok, the language/typos/grammar need more work for a documentation.

Thanks, looks the dictionary for my spell checker is not good enough.

developer_guides/uuid/index.rst
keyonjie: I assumed this will get merged after the implementation is merged, so saying "currently" here.

developer_guides/uuid/index.rst
keyonjie: um, per my understanding, the component type will be moved out of the IPC structures, but it is still used inside the firmware, let me re-organize the sentence here.

developer_guides/uuid/index.rst
keyonjie: Good learning, thanks.

developer_guides/uuid/index.rst
keyonjie: Indeed, thanks.

developer_guides/uuid/index.rst
keyonjie: The wording 'runtime' is confused here, let me correct it.

keyonjie:
@plbossart updated.

developer_guides/uuid/index.rst
keyonjie: added now.

developer_guides/uuid/index.rst
keyonjie: yes.

developer_guides/uuid/index.rst
kv2019i: Hmm, this is still a bit hard to read. How about:

UUID will be used in component creation only when support is available across the stack in FW, topology and driver. If an older FW/tplg/driver without UUID support is used, component type information is used instead.

Following table describes the firmware behaviour with different combinations of FW/tplg/driver. 

.. csv-table:: Firmware/tplg/driver ABI alignment for UUID support
   :header: "Firmware", "Topology file", "Linux driver", "Comments"
   :widths: 10, 10, 15, 40

   "no", "no", "no", "supported, no UUID"
   "no", "no", "yes", "supported, UUID 0s, FW ignore it"
   "no", "yes", "no", "supported, no UUID"
   "no", "yes", "yes", "supported, UUIDs valid, but FW ignore them, component type will be used."
   "yes", "no", "no", "supported, offset/reserved 0 passed to the FW, component type will be used."
   "yes", "no", "yes", "supported, UUID 0s, component type will be used in FW."
   "yes", "yes", "no", "supported, offset/reserved 0 passed to the FW, component type will be used."
   "yes", "yes", "yes", "supported, UUID will be supported."

lgirdwood:
@keyonjie it's probably far simpler to say tha UUID will only be used when ABI is 3.17 or greater on FW, kernel and topology otherwise type is used. This is far easier to understand than a large table.

keyonjie:
> @keyonjie it's probably far simpler to say tha UUID will only be used when ABI is 3.17 or greater on FW, kernel and topology otherwise type is used. This is far easier to understand than a large table.

Thanks, @kv2019i commented about this also, let me simplify this part.

developer_guides/uuid/index.rst
keyonjie: OK, let me remove "currently", thanks.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,129 @@
+.. _uuid:
+
+UUID Usage in SOF
+#################
+
+Why UUID Needed
+***************
+
+To develop a new audio signal processing component, we traditionally
+implement the component driver with a new unique component type
```
plbossart, line 10: implemented, it's now the past.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,129 @@
...
+UUIDs (Universally Unique Identifiers) provide a more scalable and
+collision-free way of component identification. UUIDs are used as the
+standard interface by all users of the SOF firmware, including the
+tracing subsystem, the topology .m4 files, and the Linux topology
+driver.
+
+Allocating a new UUID for the new added component is recommended, as
+the component type will be replaced by UUID in the IPC structures,
+in the future. For the whole SOF stack, the usage of the UUID should
+follow rules as below:
```
plbossart, line 31: again you are using recommended and 'should' when it's now mandatory. You have to be consistent with the language used between paragraphs.

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,129 @@
...
+`UUID wikipedia <https://en.wikipedia.org/wiki/Universally_unique_identifier>`__,
+and the value shall be declared in the firmware component driver with
+``DECLARE_SOF_RT_UUID``, for details, please refer to the API
+documentation :ref:`uuid-api`.
+
+UUID in Topology
+****************
+The same UUID shall be used in topology .m4 file for a new added
+widget (corresponding to the new added component), since we have
+implemented macro to help to handle the converting task, just use the
```
plbossart, line 45: the conversion task

developer_guides/uuid/index.rst
```diff
@@ -0,0 +1,129 @@
...
+.m4 tools/topology/m4/src.m4:
+
+.. code-block:: none
+
+   DECLARE_SOF_RT_UUID("host", host_uuid, 0x8b9d100c, 0x6d78, 0x418f,
+        0x90, 0xa3, 0xe0, 0xe8, 0x05, 0xd0, 0x85, 0x2b);
+
+Linux Topology Driver
+*********************
+The topology driver will parse the 16Bytes UUID token, append it to the
```
plbossart, line 57: 16-byte UUID token
the grammar is N-<unit> <object>, and unit is singular if I am not mistaken.

developer_guides/uuid/index.rst
plbossart: well this begs a follow-up question. how do we determine which components are used in topology files and which ones are not?

developer_guides/firmware/components/component-overview.rst
```diff
@@ -24,10 +24,12 @@ uAPI to create a component of that ``type``. It also provides an entry point
...
-collision-free way of component identification. UUIDs are currently used by
-the tracing subsystem while both the topology and driver still depend on
-``type`` to define topology and create component instances. UUIDs are
-expected to replace the ``type`` in these parts as well.
+collision-free way of component identification. UUIDs are currently used as
+the standard interface by all users of the SOF firmware, including the
+tracing subsystem, the topology .m4 files, and the Linux topology driver.
+Using the ``type`` to define topology and create component is still supported
+today, but the ``type`` could be moved out of the IPC struct in the future,
+so allocating UUID for the new added component driver is strongly recommended.
```
plbossart, line 32: again: is mandatory.

keyonjie:
> @keyonjie please revisit the language issues below (wording+grammar). You have to be consistent between shall (requirement) and should (recommendation), it's important to avoid misunderstandings.

Thanks for patience and mentoring on this, will update them next.

developer_guides/uuid/index.rst
keyonjie: Basically, the macro 'DECLARE_SOF_RT_UUID' will declare the corresponding UUID static data to the .rodata section. So, if a component(or module, or a pure driver only) is declared with macro 'DECLARE_SOF_RT_UUID' in the firmware, then its UUID should be specified in the topology. Otherwise, when declared with the macro 'DECLARE_SOF_UUID' only, it will be used for the system internal only, the tracing system will use it, but since its UUID will not located in the .rodata section, this will save the memory footprint (we need only 4 Bytes offset for each of this kind of definition).

If you grep the 'DECLARE_SOF_RT_UUID' usage in the firmware source, it is about 39 for the real use at this moment, if UUID of a component is defined in the firmware with 'DECLARE_SOF_RT_UUID', it will occupy 16 Bytes 
of the memory, no matter if the component will be used in your topology files or not.

lgirdwood:
@keyonjie btw, if you are ever unsure abour grammar then please just ask @deb-intel for review or she can also make corrections and reflow the text in English for you.

deb-intel:
@lgirdwood Thanks for your endorsement. I am happy to help with language. I really enjoy language, especially removing ambiguity for our readers. 

@plbossart is correct in that we need to make sure readers know when something is required and when something is optional or even simply encouraged. The more modern (and direct) term for "shall" is "must." If something is not required but we want readers to know that they CAN do it if they want, we can use language such as "you can," "you may," "you might want to . . .," "we recommend that you . . .," or "consider doing this . . ."

The easiest way to imply that readers must perform certain actions is to use active verbs in lists. Actions that are optional or highly recommended can be indicated as such:

1. Install this
2. Configure this
3. Download this (optional)
4. Install this (recommended although not required)

See all of the cool things we can do with language?! :-)

PR #254, opened 2020-06-05
release: update release information

kv2019i:
Remove obsolete reference to v1.3 as the latest release. It's better
not to mention the latest here at all and instead just point readers
to the GitHub release page.

Also remove text referring to linux-firmware.git and instead
refer to sof-bin repository as the official channel for binary
release packages.


lgirdwood:
Merged so we can update the correct information asap.

PR #261, opened 2020-07-06
platforms: add missing platform geminilake

cujomalainey:


cujomalainey:
Is haswell also missing or is it covered under the name broadwell?

PR #265, opened 2020-07-07
Removed references to master/slave; replaced with primary/secondary

deb-intel:

### DIFF
--- a/developer_guides/topology/topology.rst
+++ b/developer_guides/topology/topology.rst
@@ -260,7 +260,7 @@ macro:
 where:
 
 | **format**: is the SSP format ex: I2S or DSP_A or DSP_B etc
-| **mclk**: master clock in Hz
+| **mclk**: primary clock in Hz
 | **bclk**: bit clock in Hz
 | **fsync**: frame sync
 | **TDM**: TDM info including the slots, width, tx mask and rx mask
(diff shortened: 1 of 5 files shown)


developer_guides/topology/topology.rst
```diff
@@ -260,7 +260,7 @@ macro:
 where:
 
 | **format**: is the SSP format ex: I2S or DSP_A or DSP_B etc
-| **mclk**: master clock in Hz
+| **mclk**: primary clock in Hz
```
plbossart, line 263: sorry @deb-intel this change is not right. it's ok to replace master/slave core by primary/secondary core, but for clocks this is not equivalent and not a word that anyone uses.

Same for SPI primary, it's SPI master in the spec.

deb-intel: Thanks for checking, @plbossart! I noticed a few other places where I was uncertain so I did not make changes. Please let me know if I need to change back any other files. I will update the topology file right now.

plbossart: @deb-intel we should have a debate on what terms work in different contexts.

for clocks we *could* e.g. use 'main clock' for mclk, since it's really the case of a reference. likewise we could replace clock master/slave by clock provider/follower or clock owner/follower.

For things tied to hardware interfaces I don't think it helps if we rename blindly. The notion of master e.g. for SPI is configurable, e.g. you could have the SOC be a SPI slave, despite being the primary device on the board. 

I am really hoping technical writers can help refine all this with domain-specific suggestions.

PR #266, opened 2020-07-08
Replace "master" with "primary" in Mem Mangt doc

deb-intel:


lgirdwood:
@lbetlej fyi - we need to do this for FW too - lets talk about this next Tuesday.

PR #277, opened 2020-07-25
Makefile: new SOF_DOC_BUILD un-hardcodes ../sof/doc

marc-hb:
The default and recommended location for the doxygen build is a source
directory, which gets polluted by the build and incredibly tedious
to (dist)clean manually. Don't change that default and documentation
yet (baby steps) but make it at least possible to build (and clean)
doxygen outside the source directory per the usual CMake practices.

As a side-effect this makes the connection with doxygen and breathe less
implicit in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Minimal makefile for Sphinx documentation
 #
+# You can override these defaults from the command line.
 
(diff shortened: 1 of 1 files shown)


lgirdwood:
@intelkevinputnam fyi

PR #278, opened 2020-07-28
Switch internal links to :ref: 

marc-hb:
Switch internal links to `:ref:` per https://thesofproject.github.io/latest/contribute/doc_guidelines.html#internal-cross-reference-linking
    
`:ref:` internal links support offline work and don't die silently when someone changes a section name.

Fixes 37a14b1e352a, b270e65e19e15 and 2d4a0600fb9db

### DIFF
--- a/introduction/index.rst
+++ b/introduction/index.rst
@@ -206,8 +206,8 @@ Which Xtensa toolchains does SOF currently support?
      documented in the getting started guide. These must be built from
      source. For instructions, refer to the following:
 
-     - `Build toolchains from source <https://thesofproject.github.io/latest/getting_started/build-guide/build-from-scratch.html#step-2-build-toolchains-from-source>`_
-       in the Getting Started Guide for building SOF from scratch
+     - :ref:`build-toolchains-from-source` in the Getting Started Guide
+       for building SOF from scratch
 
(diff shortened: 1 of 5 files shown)


introduction/index.rst
```diff
@@ -245,4 +245,4 @@ Cadence xt-xcc or Cadence xt-clang?
 
 Is support for other toolchains forthcoming?
   Going forward, we would like to support the LLVM C compiler. Patches are
-  welcome.
\ No newline at end of file
+  welcome.
```
ktrzcinx: Here is some unintended changes, I guess

marc-hb: Thanks @ktrzcinx for the review. The couple newlines changes are intended.

While github doesn't show it well, the last line in this file had no newline at the end and this commit adds it. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

The only other whitespace changes are breaking down paragraphs that were entirely on one line, this will make `git diff` usable in the future. I did this only for the paragraphs I was already fixing anyway.

marc-hb: Correction: one of the long paragraphs I broke down has no other, real change, my bad. I don't think it's worth it but I can undo this one of you insist .

ktrzcinx: I won't insist.

PR #281, opened 2020-08-14
Created new fuzzy dir and added two related docs to it.

deb-intel:
Made grammatical edits to testbench fuzzy doc.

Created a new directory called Fuzzing and added the two fuzzy files to it. This is to keep the top-level items under Developer Guides to a minimum.
--Fuzzing in Docker
--Build a Fuzzing Testbench with AFL


deb-intel:
@ymdatta Please review my changes to the Fuzzy documentation when you have a moment. Thanks!

developer_guides/fuzzing/testbench_afl_fuzzing.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   $HOME/work/
+
+
+Build a testbench with AFL instrumentation
+******************************************
+
+According to AFL's `README <https://github.com/google/AFL/blob/master/README.md>`_, AFL is a "brute-force fuzzer coupled with an exceedingly
+simple but rock-solid instrumentation-guided genetic algorithm." Add
+instrumentation to the code before running a fuzzer on it in order to get
```
ymdatta, line 25: I feel here we need to specify it's really important to add instrumentation.

When i ran fuzzer on some parts of code without instrumentation i didn't get any results. But on adding the instrumentation, there were some results. So, i feel it's a bit necessary to say the reader that it's important to instrument before running the fuzzer.

developer_guides/fuzzing/testbench_afl_fuzzing.rst
```diff
@@ -0,0 +1,88 @@
...
+input files. This is the ``testcase_dir`` in the above command.
+
+Since you are fuzzing the testbench, the ``program`` here is testbench.
+
+``params`` are the different parameters of the program apart from the input
+file.
+
+``@@``: Each file from ``testcase_dir`` is substituted in place of this.
+As AFL continues to run, newly-generated testcases are placed in
+``testcase_dir``, and AFL is run again with those testcases.
```
ymdatta, line 58: I am nitpicking here.

Do you think it will be better to write it as "and AFL in its further iterations runs with these newly generated testcases" instead of "and AFL is run again with those testcases"?

developer_guides/fuzzing/testbench_afl_fuzzing.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Example
+*******
+
+Use AFL to fuzz the volume component of the testbench. Use topology files as inputs and place the topology files of volume components in an ``inputs`` directory:
```
ymdatta, line 63: Again a small nitpick.

I think it's better if we write 'Using AFL to fuzz the volume component of the testbench' on a separate line at the beginning. As this helps the reader in understanding what this example is trying to do.

developer_guides/fuzzing/testbench_afl_fuzzing.rst
deb-intel: Updated to:
You must add instrumentation to the code before running a fuzzer in order to get potentially useful results; otherwise, you might not get any results.

*Note that I changed it to "potentially useful" results as opposed to "good" because we really cannot guarantee that the results they get will be good.

developer_guides/fuzzing/testbench_afl_fuzzing.rst
deb-intel: Good advice

deb-intel: Updated!

deb-intel:
Thanks for your review, @ymdatta. You were not nitpicking; your suggestions improved this document.

ymdatta:
LGTM!

PR #295, opened 2020-10-15
getting_started: add vital kernel links

marc-hb:
Due to the lack of these links, the very first time I read this section
I assumed there was no other "kernel HOWTO" than this section elsewhere
in sof-docs and I did not search for anything else. So I did obviously
not find the pages linked to by this commit and I tried to figure
everything on my own instead which probably caused my biggest waste of
time when starting to work on SOF. No exaggeration.

I'm not sure why the ktest section is placed under "Setting up SOF on
hardware", maybe a different location for it would have helped me
too. Anyway adding a couple links is much easier than moving sections
around and a getting started guide is a good place for links to more
advanced documentation in any case. This getting started guide has
surprisingly few such links to more advanced topics.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -171,6 +171,8 @@ extlinks = {
        (SOF_GIT + '/sof/tree/master/%s', ""),
     'git-sof-docs-mainline':
        (SOF_GIT + '/sof-docs/tree/master/%s', ""),
+    'git-sof-kconfig':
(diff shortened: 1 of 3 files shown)


deb-intel (approved):
Nice additions!

lgirdwood (approved):
Thanks for fixing @marc-hb

PR #299, opened 2020-10-20
Logger filtering extra info

ktrzcinx:
Add few extra information about trace filtering system limitations.

### DIFF
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -214,6 +214,11 @@ It may be done by:
 Logger trace filtration affect only traces sent after filter setup, so traces
 already stored on kernel side will not be affected.
 
+Filters are setup incrementally, so when logger will be run-up twice with
+different settings, then filters from first run will not be restored to default
+state but can be replaced by new one.
+To reset filters to default state, firmware reset is needed.
+
 Detailed description
 --------------------
 
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -211,6 +211,9 @@ It may be done by:
   
     cat /sys/devices/pci0000\:00/0000\:$(lspci -nn | grep "audio controller" | awk '{print $1;}')/power/runtime_status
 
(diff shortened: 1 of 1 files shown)


developer_guides/debugability/logger/index.rst
```diff
@@ -194,6 +194,10 @@ A similar example may be prepared for components on a particular pipeline:
   To track verbose message use firmware build with selected "Trace verbose"
   option under "Trace" menu.
 
+Active trave filters are stored in firmware runtime memory, so after firmware
+restart (eg. after power gating in sleep mode) filters settings will be reset.
+Consider disabling power gating during your debug session.
```
lgirdwood, line 199: We should have the instruction for disabling runtime PM here. @plbossart @kv2019i is there a one line command that @ktrzcinx can insert here.

lgirdwood (changes requested):
We need to include the Linux command line for disabling the SOF driver runtime PM. i.e. this guide should be a one stop shop for tracing

developer_guides/debugability/logger/index.rst
xiulipan: @ktrzcinx I now see the debugfs filter is read only. Can we make it to be read/write and restore the filter when we send CTX_RESTORE?

ktrzcinx: @lgirdwood I've added cmd to disable PM

@xiulipan I would be nice solution to restore filter setup, but there are a one problematic issue. FW doesn't store default filter settings, so to restore them, FW restart is needed, so in situation when logger is run-up twice with different filter setup without FW reboot, then combination of filters will be used - it already works like this, but now user must be aware of this, after introducing this automat, he won't be.

xiulipan: @ktrzcinx Let's try to create a discussion about this feature update. I think end user will not use this, we can collect feedback from developer.

lgirdwood (approved):

deb-intel (approved):
Will make minor grammatical edits in a separate PR.

PR #309, opened 2020-11-12
ktest: add grub-reboot cheat sheet

marc-hb:
I found all `grub-reboot` guides either too long or too short or confusing
or all of the above :-)

### DIFF
--- a/getting_started/setup/setup_ktest_environment.rst
+++ b/getting_started/setup/setup_ktest_environment.rst
@@ -59,15 +59,71 @@ working kernel if your changes fail to boot.
 
    # Use your text editor of choice.
    sudo emacs /etc/default/grub
+   sudo update-grub
 
 Add ``GRUB_DISABLE_SUBMENU=y`` to the end and save.
+Sub-menus confuse ktest.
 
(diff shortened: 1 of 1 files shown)


getting_started/setup/setup_ktest_environment.rst
```diff
@@ -59,15 +59,71 @@ working kernel if your changes fail to boot.
 
    # Use your text editor of choice.
    sudo emacs /etc/default/grub
+   sudo update-grub
```
fredoh9, line 62: Should we put what we edit in /etc/default/grub in between?

fredoh9 (commented):
Thanks Marc, I liked the details about grub-reboot. This will be the bible of ktest.

getting_started/setup/setup_ktest_environment.rst
marc-hb: Logically yes but that would consume two  `code-block`s for two lines, a bit too verbose IMHO. I'm assuming the reader's look-ahead buffer is at least 2 lines long...

deb-intel (approved):
Nicely worded!

PR #310, opened 2020-11-12
ktest: add cleaning up old kernel modules

fredoh9:
Under current ktest, user needs to clean up old kernel modules manually.

Signed-off-by: Fred Oh <fred.oh@linux.intel.com>


plbossart (approved):

marc-hb (approved):
I find this added information especially useful considering the sample ktest.conf above relies on an untold but useful no-kernel-version "hack" that is incompatible with update-initramfs and causes a warning in update-grub:

```
Generating grub configuration file ...
dpkg: warning: version 'test' has bad syntax: version number does not start with digit
Found linux image: /boot/vmlinuz-test
Found initrd image: /boot/initrd.img-test
```

Maybe "-test" should include a dummy version number? I digress sorry.

PR #313, opened 2020-12-15
ktest: un-hardcode $LOCALVERSION

marc-hb:
Among other meaninglessness sins, 'test' is not SOF- or even audio-
specific and ungreppable. Don't change it the sample
configuration (yet?)  but at least give the option to change it.

This requires dropping the minus sign from '-test' example because
update-grub uses minus as a separator and ktest fails with:

  Could not find ''Ubuntu, with Linux -test''

Also add some comments explaining why things work despite the hack where
/lib/modules/x.y.z-LOCALVERSION does not match /boot/vmlinuz-$LOCALVERSION


lgirdwood (approved):
Tech content is good.

deb-intel (approved):

PR #314, opened 2020-12-17
build-guide: Add note for using external key in signing

juimonen:
Add info how the set environment variable for using external key in signing.

Signed-off-by: Jaska Uimonen <jaska.uimonen@intel.com>


deb-intel (approved):

lgirdwood (approved):

marc-hb (commented):
This new section is a copy/paste/diverge of the `-h` usage message, I'd like it to at least mention that there is a usage message which is much, much less likely to get out of sync with the code because it's implemented in the script itself.

PS: can we have at least a day or two between submission and merge? This didn't seem like an urgent fix.

juimonen:
@marc-hb yes exactly added you here because wanted to get your comments... but let's refine this. I got external request to update this info somewhere in the docs, we would for sure need more info on rimage (its own section in sof-docs is quite limited).

PR #320, opened 2021-01-26
Convert .travis.yml to github actions

marc-hb:
"deploy" steps not converted yet. I have never seen them run, are they
still in use?

Either way this first and most important part can be merged already.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,107 @@
+---
+# Tools that can save round-trips to github and a lot of time:
+#
+# yamllint -f parsable pull_request.yml
+# pip3 install ruamel.yaml.cmd
+# yaml merge-expand pull_request.yml exp.yml &&
+#    diff -w -u pull_request.yml exp.yml
+#
+# github.com also has a powerful web editor that can be used without
+# committing.
+
+name: Pull Requests
+
+# yamllint disable-line rule:truthy
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# As of January 2021, no YAML anchors :-(
+env:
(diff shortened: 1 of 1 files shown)


paulstelian97 (approved):
Minor comment: In the places where you have multiple steps, they always work in sequence? I don't see any dependencies and I don't yet know if you can have them syntactically.

xiulipan (approved):
This is exactly same with the Travis jobs, also had some confusing job name like **Supported scripts/requirements.txt** **PIP_IGNORE_INSTALLED=0 requirements-lax.txt**

Also we have a TODO: the deploy part.

PS: what is the LAX means here?

lgirdwood (approved):

lgirdwood:
@marc-hb I assume we should see the Travis "pending" with this PR since it's being moved to GH actions ?

marc-hb:
> Minor comment: In the places where you have multiple steps, they always work in sequence?

Yes: steps are sequential and jobs are concurrent.

> This is exactly same with the Travis jobs, also had some confusing job name like Supported scripts/requirements.txt PIP_IGNORE_INSTALLED=0 requirements-lax.txt
> PS: what is the LAX means here?

The answer to both questions is in the heavily commented `scripts/requirements-lax.txt`

I agree these job names are not great, suggestions welcome... for a future PR? We need this one urgently because:

> @marc-hb I assume we should see the Travis "pending" with this PR since it's being moved to GH actions ?

No, it's just because Travis has been moving most of their systems from travis.org to travis.com to travis.org is overloaded all the time (hence this PR)
https://www.traviscistatus.com/#week

marc-hb:
@deb-intel , @intelkevinputnam , are the "deploy" jobs in .travis.yml still in use?

xiulipan:
@marc-hb We need that job to automatically upload the build result to https://github.com/thesofproject/thesofproject.github.io
As Travis now get delay in job run, it means we will also get delay to auto upload build result to https://github.com/thesofproject/thesofproject.github.io
I believe GitHub action should have better support for such kind of usage? I can have a look first

PR #322, opened 2021-02-01
Addition of files used to run Woke word checker.

deb-intel:

Includes two files:
.wokeignore lists files to ignore when running Woke.
rules-woke.yaml is the customized file that must be run to accurately identify and correct non-inclusive terms.
    This file is called by woke -c ./rules-woke.yaml

### DIFF
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,0 +1,20 @@
+# The following files can be ignored when running Woke.
+
+rules-woke.yaml
+conf.py
+*.pu
+tox.ini
+Makefile
+.travis.yml
+.github/workflows/*.yml
+index.rst
+release.rst
+introduction/index.rst
+maintainers/merge_rights.rst
+contribute/process/bug-tracking.rst
+getting_started/setup/setup_up_2_board.rst
+developer_guides/virtualization/files/q-v6.sh
+developer_guides/topology/topology.rst
+developer_guides/fuzzing/testbench_afl_fuzzing.rst
+
+
--- a/rules-woke.yaml
+++ b/rules-woke.yaml
@@ -1,0 +1,39 @@
+# Use this customized file to ensure that non-inclusive language is identified
+# and corrected. Keep it in the top level of the sof-docs directory.
+# Customization and improvement of this file is ongoing.
+# The following command calls this file:
+#  woke -c ./rules-woke.yaml
+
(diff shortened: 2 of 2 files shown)


lgirdwood (approved):

lgirdwood:
@xiulipan ready when you are.

rules-woke.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - main for firmware-related topics and provider for hardware-related topics
+
+  - name: blacklist
+    terms:
+      - blacklist
+      - black-list
+    alternatives:
+      - blocklist
+
+  - name: whitelist
```
xiulipan, line 34: I see duplicate `whitelist` here.

xiulipan (changes requested):
I found some duplicated rules.

xiulipan:
@deb-intel @lgirdwood I am fine with this configure files, please merge if you solve the duplicated rules above. Then I will work on this base for the woke.

xiulipan:
@deb-intel Good news, the public market action can work with extend rules. see https://github.com/xiulipan/sof-docs/pull/1#discussion_r568308045 for detail.
I will try to modify the wokeignore to avoid such issues. But it prove that our customized rules is working.

.wokeignore
```diff
@@ -0,0 +1,20 @@
+# The following files can be ignored when running Woke.
+
+rules-woke.yaml
+conf.py
+*.pu
+tox.ini
+Makefile
+.travis.yml
+.github/workflows/pull-request.yml
```
xiulipan, line 9:
```suggestion
.github/workflows/*.yml
```
I think we will add more code checked as actions later.

rules-woke.yaml
deb-intel: Nice catch. Corrected

.wokeignore
deb-intel: Modified.

xiulipan (approved):
LGTM.
PS: do we need this for other project?

deb-intel:
@xiulipan Are you referring to a particular project?

lgirdwood:
> @xiulipan Are you referring to a particular project?

It's the source code for FW and tools etc, but we can do this after CNY when @xiulipan is back.

PR #321, opened 2021-01-28
Enable sof-docs deploy in GitHub Actions

xiulipan:
Disable Travis to avoid conflict.
Optimize and simplify GitHub Actions
Add new deploy job for publish branch

test with my own forked repo
- deploy repo commit
https://github.com/xiulipan/xiulipan.github.io/tree/master
- Action log
https://github.com/xiulipan/sof-docs/actions/runs/518058071
- Final HTML
https://xiulipan.github.io/latest/index.html

TODO: we may need someone with admin to sof-docs to create a personal access token
and set a variable **ACTIONS_DEPLOY_KEY** as encrypted secrets for a repository. 

guide are attached:
https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
(the only needed access is **repo/public_repo**)
https://docs.github.com/en/actions/reference/encrypted-secrets

PR #323, opened 2021-02-02
optimization for GitHub Actions

xiulipan:
Did some optimization for GitHub Actions:
- change the action name
- Add and use the local actions for duplicated steps.
- simplify ENV setup steps, merge single steps into one step to give more clear result view.

### DIFF
--- a/.github/actions/build-sof-doc.sh
+++ b/.github/actions/build-sof-doc.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+# build doc from sof
+git clone https://github.com/thesofproject/sof
+cmake -GNinja -S sof/doc -B _build_doxy
+
+# TODO: change the (bad) default value for SOF_DOC_BUILD in
+# sof-docs/Makefile and remove this command line override
+# build sof-doc
+make html VERBOSE=1 SOF_DOC_BUILD=_build_doxy
+du -shc _build*/*
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -44,18 +44,8 @@ jobs:
       - name: 'pip install -r scripts/requirements.txt'
         run: pip install -r scripts/requirements.txt
 
-      # No YAML anchors; this is copy/paste
-      - name: configure SOF doc
-        run: |
-          git clone https://github.com/thesofproject/sof
-          cmake -GNinja -S sof/doc -B _build_doxy
-
-      # TODO: change the (bad) default value for SOF_DOC_BUILD in
-      # sof-docs/Makefile and remove this command line override
-      - name: build
-        run: |
-          make html VERBOSE=1 SOF_DOC_BUILD=_build_doxy
-          du -shc _build*/*
+      - name: build doc
+        run: ./.github/actions/build-sof-doc.sh
 
   build-lax:
 
@@ -95,13 +85,5 @@ jobs:
            # We don't want plantUML to raise the contribution bar
            sed -i -e 's/^\(plantuml_output_format *=\).*/\1 "none"/' conf.py
 
-      # No YAML anchors; this is copy/paste
-      - name: configure SOF doc
-        run: |
-          git clone https://github.com/thesofproject/sof
-          cmake -GNinja -S sof/doc -B _build_doxy
-
-      - name: build
-        run: |
-          make html VERBOSE=1 SOF_DOC_BUILD=_build_doxy
-          du -shc _build*/*
+      - name: build doc
+        run: ./.github/actions/build-sof-doc.sh
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -9,7 +9,7 @@
 # github.com also has a powerful web editor that can be used without
 # committing.
 
-name: Pull Requests
+name: GitHub Actions
 
 # yamllint disable-line rule:truthy
 on:
@@ -27,9 +27,9 @@ env:
 
 jobs:
 
-  supported-reqs:
+  build-basic:
 
-    name: 'Supported scripts/requirements.txt'
+    name: "build with requirements.txt"
     runs-on: ubuntu-20.04
 
     steps:
@@ -57,9 +57,9 @@ jobs:
           make html VERBOSE=1 SOF_DOC_BUILD=_build_doxy
           du -shc _build*/*
 
-  lax:
+  build-lax:
 
-    name: "PIP_IGNORE_INSTALLED=0 requirements-lax.txt"
+    name: "build with requirements-lax.txt"
     runs-on: ubuntu-20.04
     # Makefile downgrades the Sphinx warnings, they are not errors any more
     env: {LAX: 1}
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -35,14 +35,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: apt-get install
-        run: sudo apt-get -y install $ubuntu_base_deps
-
-      - name: PATH += .local/bin
-        run: echo "$HOME/.local/bin" >> $GITHUB_PATH
-
-      - name: 'pip install -r scripts/requirements.txt'
-        run: pip install -r scripts/requirements.txt
+      - name: setup env with requirements.txt
+        run: |
+          sudo apt-get -y install $ubuntu_base_deps
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          pip install -r scripts/requirements.txt
 
       - name: build doc
         run: ./.github/actions/build-sof-doc.sh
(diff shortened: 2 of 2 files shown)


xiulipan:
@marc-hb @lgirdwood @deb-intel Moved from #321 all requirement is addressed here

lgirdwood (approved):

marc-hb (changes requested):
As mentioned in previous #321 , there are good things here and others I'm not OK with.

Let's first focus on the finishing the migration thanks to your #324 then we can get back to this.

.github/workflows/pull-request.yml
```diff
@@ -27,81 +27,48 @@ env:
...
-
-      # No YAML anchors; this is copy/paste
-      - name: configure SOF doc
+      - name: setup env with requirements.txt
         run: |
-          git clone https://github.com/thesofproject/sof
-          cmake -GNinja -S sof/doc -B _build_doxy
+          sudo apt-get -y install $ubuntu_base_deps
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          pip install -r scripts/requirements.txt
```
marc-hb, line 42: Collapsing all these steps means more time spent looking for an error when something is wrong in a PR. What is the purpose?

.github/actions/build-sof-doc.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# build doc from sof
+git clone https://github.com/thesofproject/sof
+cmake -GNinja -S sof/doc -B _build_doxy
+
+# TODO: change the (bad) default value for SOF_DOC_BUILD in
+# sof-docs/Makefile and remove this command line override
+# build sof-doc
+make html VERBOSE=1 SOF_DOC_BUILD=_build_doxy
+du -shc _build*/*
```
marc-hb, line 10: Thanks for moving this to a script

.github/workflows/pull-request.yml
```diff
@@ -27,81 +27,48 @@ env:
...
-
-      - name: config tweaks
+      - name: setup env with requirements-lax.txt
         run: |
-           # sphinx-build < 1.7 doesn't support "auto" and we're not sure
-           # all default modules are "thread-safe"
-           sed -i -e '/SPHINXBUILD/ s/-j *auto/-j 1/' Makefile
-           # We don't want plantUML to raise the contribution bar
-           sed -i -e 's/^\(plantuml_output_format *=\).*/\1 "none"/' conf.py
+          sudo apt-get update
```
marc-hb, line 58: As already mentioned in #321, if we add `apt-get update` here then this job will be slow forever. The `wget` is slow too but it will hopefully fail one day in the future. When it starts failing we remove it and we are fast again. I should have explained this in a comment.

.github/workflows/pull-request.yml
```diff
@@ -27,81 +27,48 @@ env:
...
+          sudo apt-get -y install $ubuntu_base_deps
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          PIP_IGNORE_INSTALLED=0 pip install -r scripts/requirements-lax.txt
```
marc-hb, line 61: The comment "apt-get instead of pip" was here for a reason. That reason is explained in `scripts/requirements-lax.txt`.

marc-hb (changes requested):
This PR reverts a number of conscious, commented design decisions without trying to understand them. Strange.

lgirdwood:
@xiulipan @marc-hb I need to merge something tomorrow after any review changes as we are blocking publication and v1.7-rc1 has just been tagged. We can deal with any optimisations after CNY as long as we can track them. The Travis job is used for publication so this needs to be a GH action now.

marc-hb:
The publication (and final) feature has been migrated from Travis to Github Actions and merged in #324 . Everything should work now, this #323 is pure refactoring and it should not block anything.

.github/workflows/pull-request.yml
lgirdwood: This looks like a one time task ? Can we optimise this later for only running once ?

lgirdwood:
> The publication (and final) feature has been migrated from Travis to Github Actions and merged in #324 . Everything should work now, this #323 is pure refactoring and it should not block anything.

@deb-intel @intelkevinputnam can you confirm this work for you now or do we still have a blockage somewhere else ?

.github/actions/build-sof-doc.sh
marc-hb: This is just a fancier `ls` to show what was just built and make sure it looks roughly as expected. This will show any mistake and or confusion between with `_build`, `_build_doxy`, `SOF_DOC_BUILD` etc.

intelkevinputnam:
Yep. With #324 in it looks good to go.

.github/workflows/pull-request.yml
xiulipan: @marc-hb Could you help to update about how to use `scripts/requirements-lax.txt`. I have tried to read this file for many times but could not find what exact steps I need if I would run it locally.
@deb-intel @intelkevinputnam Any idea about what is the `scripts/requirements-lax.txt` and how will we use it?

.github/workflows/pull-request.yml
xiulipan: I just want to reproduce what we will do locally from the document we shared. But it seems I still get confused about how to use it.

xiulipan:
@marc-hb @lgirdwood I am fine with the current solution as long as it can work.
This optimization is based on the following points summarized from my personal understanding and experience with playing GitHub Action and review demo and public script:

-  merge some related CMD into one job
    each CDM to be a job will make the whole workflow looks redundant and miss the key point for the test.
-  Do not use special setup steps for GitHub 
    We should make the environment set up process exactly like what we shared in our README and we will do locally. (PS: I do not see apt update **slower** then the wget way. You can check the actions history logs and I have also shared the detail logs in previous discussion)

.github/workflows/pull-request.yml
deb-intel: @xiulipan I think this was information created by @intelkevinputnam for purposes of doc publishing (I do not know the specifics).

lgirdwood:
> @marc-hb @lgirdwood I am fine with the current solution as long as it can work.
> This optimization is based on the following points summarized from my personal understanding and experience with playing GitHub Action and review demo and public script:
> 
> * merge some related CMD into one job
>   each CDM to be a job will make the whole workflow looks redundant and miss the key point for the test.
> * Do not use special setup steps for GitHub
>   We should make the environment set up process exactly like what we shared in our README and we will do locally. (PS: I do not see apt update **slower** then the wget way. You can check the actions history logs and I have also shared the detail logs in previous discussion)

Ok, lets revisit after CNY, we can see how things work atm. Thanks @xiulipan

.github/workflows/pull-request.yml
marc-hb: > I have tried to read this file for many times but could not find what exact steps I need if I would run it locally.

What happens when you just run the commands here in this workflow?

marc-hb (changes requested):
> merge some related CMD into one job
> each CDM to be a job will make the whole workflow looks redundant and miss the key point for the test.

I assume you meant "step" instead of "job".

I don't understand what "redundant" means here. As I wrote, having more steps means faster failure identification because no need to open and scan all the logs. I spent a fair amount of time reading https://docs.github.com/en/actions and I never saw more than 2 commands per step there.

The "key point" is expressed in the job names which I agree are bad right now, thanks for offering some better job names suggestions.

> Do not use special setup steps for GitHub. We should make the environment set up process exactly like what we shared in our README and we will do locally

Because of "Python DLL hell" there are unfortunately two alternative setups to build the docs: one officially supported but strict and another more convenient for some people.  If you're interested, read the unfortunately long explanations in commit e987dcf322 and PR #198. If you're not interested, just use the setup you prefer and don't go and remove the other setup you don't like without understanding why it exists. Refactoring anything requires a complete understanding, otherwise and if it works just leave it alone.

> PS: I do not see apt update slower then the wget way.

This is off-topic. You keep missing the point and I ran out of ways to rephrase it sorry.

.github/workflows/pull-request.yml
xiulipan: All I do understand in previous script is about
```
           sed -i -e '/SPHINXBUILD/ s/-j *auto/-j 1/' Makefile
           # We don't want plantUML to raise the contribution bar
           sed -i -e 's/^\(plantuml_output_format *=\).*/\1 "none"/' conf.py
```

I could not find them in any other part expect .travis.yml do we also need them locally?

xiulipan:
@marc-hb I may not understand what we are doing with lax part here and it seems no other one can help me to understand here. I will close this PR, please take any optimization you think may need and send a new PR.

.github/workflows/pull-request.yml
marc-hb: > If we add apt-get update here then this job will be slow forever. The wget is slow too but it will hopefully fail one day in the future. When it starts failing we remove it and we are fast again

I was wrong, Github Actions require `apt-get update` before `apt-get install` because unlike normal Ubuntu systems they don't `apt-get update` in the background. Longer story in https://github.com/actions/virtual-environments/issues/ 1757 (whitespace inserted to avoid spamming 1757)

PR #343, opened 2021-04-22
components: update the device state machine diagram

ranj063:
Update the component device state machine diagram
to remove the transitions during the XRUN trigger.

Signed-off-by: Ranjani Sridharan <ranjani.sridharan@linux.intel.com>


lgirdwood (approved):
@juimonen fyi

PR #344, opened 2021-04-29
developer_guides: add DSP cores management doc

keyonjie:
Add documentation about DSP cores management.

Signed-off-by: Keyon Jie <yang.jie@linux.intel.com>


developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
+scheduler widget's core).
+
+Each component (normal driver widget) can have its own task also, it can be
+specified to run on a specified DSP core, this is not neccesary to be the same
+one with the pipeline one.
+
+
+How to manage the DSP cores in the driver?
+******************************************
+
```
lyakh, line 35: do we need a ref-count for the main core? We can do that for symmetry, but I'd only let it hit 0 when switching the DSP off for various suspend cases. So, the difference is, that all (secondary cores) are ref-counted for any audio-streaming, that they're needed for, whereas the main core is kept on until suspend. Maybe just increment its use-count by one as an "OS core."

And then begin by explaining that: that we want to power secondary cores on and off when using them for audio.

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
+  via IPC component_new.
+- decrease the refcount via _put() after a widget is freed, in widget_free().
+- power on a DSP core at the point the refcount is changing from 0 to 1, and
+  power off it at 1 to 0.
+
+How to power on/off a DSP core?
+*******************************
+
+To power on a DSP core, we need to
+
```
lyakh, line 49: we first have to explain, that some cores are "host-managed" and some are "DSP-managed."

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
+The life cycle of a DSP core
+****************************
+
++-------------+---------+--------------+---------------------------+----------------------------+----------------------+
+| Cores/stage | booting	|    booted    |	widget setup	   |       widget free		|	suspend        |
++=============+=========+==============+===========================+============================+======================+
+| Master Core |  ON	|   ON(ref=1)  |   ON(_get(),ref=2,3,...)  |  ON(_put(), ref=...2,1)	|   OFF(_put(), ref=0) |
++-------------+---------+--------------+---------------------------+----------------------------+----------------------+
+| Slave Cores |	 ON	|   OFF(ref=0) |   ON(_get(),ref=1,2,...)  | ON/OFF(_put(), ref=...1,0)	|      OFF (already)   |
```
lyakh, line 70: We use "primary" and "secondary" I think?

lyakh (changes requested):

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
++-------------+---------+--------------+---------------------------+----------------------------+----------------------+
+
+.. note:: The booting stage could be diverse, some platforms has requirement
+   that we need to power on all cores while some others only requires core 0.
+   To simplify this, we just treat the booting stage as the initialization
+   stage of the DSP cores status, that is, after booted, master is intialized
+   to ON(ref=1) and slave ones are OFF(ref=0).
+   At the same time, the IPC comunication is not set up yet in this booting
+   stage, so we don't use the _get/_put() helpers at this stage, just configure
+   registers in the driver side to do this initialization.
```
kv2019i, line 80: Not sure about this. I mean, we could also push this logic to DSP driver implementation. E.g. on Intel TGL, the DSP implementation knows that core 1 needs an IPC, while core 0 can be powered directly. On Intel ICL, cores are powered directly. This logic should apply in the boot phase as well (or if not, it's a DSP specific difference that should be hidden in the DSP specific implementation).

kv2019i (changes requested):
Thanks for starting this! The main open I have is whether the concept of "host managed" is made a generic property of SOF, or something that is an implementation detail of a specific DSP. I.e. Intel TGL and ICL require different flows on this and currently this is not exposed to SOF core.

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
+.. _dsp_core:
```
ranj063, line 1: @keyonjie thanks for starting this document but I think we should focus more on how the cores should be managed instead of how the users (ie widgets/pipelines etc) use them.

So I would make the document focus first on how the cores should be managed ie with ref_counting to account for the fact that core must be powered on until the last user releases the ref_count. And then follow up it with the example of how topology components use ref_counting.

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
+  DSP core tearing down in the FW side,
+- power off it in the driver side (if the core is host manged) usually via
+  register setting.
+
+The life cycle of a DSP core
+****************************
+
++-------------+---------+--------------+---------------------------+----------------------------+----------------------+
+| Cores/stage | booting	|    booted    |	widget setup	   |       widget free		|	suspend        |
++=============+=========+==============+===========================+============================+======================+
```
ranj063, line 67: This is very dynamic pipeline and topology specific. What we want is more of how a user (not just topology) should use the multi-core implementation. ie what APIs to use for requesting a core and releasing it when no longer needed. For example, the FW booting is simply a user rather than a stage in the DSP core's life. 

we are tying 2 separate concepts here and I think our main focus should be how cores are managed by the driver and firmware rather than how dynamic pipelines should use the cores.

developer_guides/dsp_core/index.rst
ranj063: I agree. I dont think we power on all cores during boot. We only power up the primary ON on ICL+.

We should move the Intel-spefic multi-core implementation to a separate section

keyonjie: > I agree. I dont think we power on all cores during boot. We only power up the primary ON on ICL+.
> 
> We should move the Intel-spefic multi-core implementation to a separate section

We have logic like this in cl_dsp_init():

```
	/* step 1: power up corex */
	ret = snd_sof_dsp_core_power_up(sdev, chip->host_managed_cores_mask);
...
	/* step 5: power down cores that are no longer needed */
	ret = snd_sof_dsp_core_power_down(sdev, chip->host_managed_cores_mask &
					  ~(chip->init_core_mask));
```

developer_guides/dsp_core/index.rst
keyonjie: > @keyonjie thanks for starting this document but I think we should focus more on how the cores should be managed instead of how the users (ie widgets/pipelines etc) use them.
> 
> So I would make the document focus first on how the cores should be managed ie with ref_counting to account for the fact that core must be powered on until the last user releases the ref_count. And then follow up it with the example of how topology components use ref_counting.

That's true, but some context about how and why a widget/component should take a ref_counting of the cores could help to understand the full picture, IMHO.

developer_guides/dsp_core/index.rst
keyonjie: > This is very dynamic pipeline and topology specific. What we want is more of how a user (not just topology) should use the multi-core implementation. ie what APIs to use for requesting a core and releasing it when no longer needed. For example, the FW booting is simply a user rather than a stage in the DSP core's life.

Yep, that sounds more like a user guide for driver writing, sounds good.

> 
> we are tying 2 separate concepts here and I think our main focus should be how cores are managed by the driver and firmware rather than how dynamic pipelines should use the cores.

I added this table, with intention to provide a full picture about all the users who will touch the DSP cores power change, and a typical sequence about how they interact with each other.

developer_guides/dsp_core/index.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Context -- pipeline, component, and DSP core
+********************************************
+
+Driver pipeline and widget
+==========================
+
+A pipeline usually contains one scheduler widget (or named 'pipeline widget in
+the FW') + several other widgets, all these widgets are 'swidget's in the
+driver side.
```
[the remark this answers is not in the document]
marc-hb: I with I had found such a nice, short and high-level introduction to pipelines and components concepts much earlier. It deserves to be much more visible than buried in a multicore document and replaced here by a link to it.

greg-intel (commented):
Information looks very helpful. 
Thanks for this.

lgirdwood:
@keyonjie if this is technically correct then perhaps @anton-intel could help with the English grammar and cleanup ? Ok for you guys ?

marc-hb:
I could be wrong but I think both @anton-intel and @deb-intel prefer to make English fixes on merged PRs. I know because I missed some corrections :-) https://github.com/thesofproject/sof-docs/pull/383#issuecomment-974298377

We have no automated publication of the tip of the main branch right now so having imperfect English in some commits in the history is fine by me.

anton-intel:
@marc-hb, True, I'll create a new PR with the edits.

PR #345, opened 2021-05-19
conf.py: Add support for sphinx 4.0+

ujfalusi:
add_stylesheet is removed with sphinx 4.0 and this will cause error in
conf.py
Fix it by trying to use the new add_css_file introduced in 1.8 and fall
back to add_stylesheet for older sphinx version.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -185,7 +185,10 @@ html_static_path = ['static']
 def setup(app):
 # add_stylesheet() was renamed to add_css_file() in sphinx 1.8 released
 # in September 2018. add_stylesheet() will be removed in sphinx 4.0
-   app.add_stylesheet("sof-custom.css")
+    try:
+        app.add_css_file('sof-custom.css')
+    except AttributeError:
+        app.add_stylesheet('sof-custom.css')
 
 # Custom sidebar templates, must be a dictionary that maps document names
 # to template names.


lgirdwood (commented):
@intelkevinputnam this needs your inputs.

conf.py
```diff
@@ -185,7 +185,10 @@
 def setup(app):
 # add_stylesheet() was renamed to add_css_file() in sphinx 1.8 released
 # in September 2018. add_stylesheet() will be removed in sphinx 4.0
-   app.add_stylesheet("sof-custom.css")
+    try:
+        app.add_css_file('sof-custom.css')
+    except:
```
marc-hb, line 190: Can you please catch a more specific exception? It should be printed when it fails.

marc-hb (changes requested):
```
pylint --disable=C conf.py
conf.py:190:2: W0702: No exception type(s) specified (bare-except)
```
https://www.flake8rules.com/rules/E722.html
... otherwise looks good.

conf.py
ujfalusi: @marc-hb, sure, let me take a look. But I don't think there is a point to print anything.
https://thesofproject.github.io/latest/contribute/process/docbuild.html states that we need Sphinx 1.7.5 or newer. add_css_file() is added with 1.8.0, it is not an error if we need to fall back to add_stylesheet.

off topic: Do we want to fix the other existing warnings from pylint?

ujfalusi: @marc-hb, pushed a new version to catch the AttributeError exception which will be raised when the add_stylesheet is not available (pre Sphinx 1.8.0).

marc-hb: > But I don't think there is a point to print anything.

I don't think that either, this was just to help you find `AttributeError`.

> off topic: Do we want to fix the other existing warnings from pylint?

Maybe some of them in a different PR why not.

marc-hb (approved):

PR #350, opened 2021-05-28
Toolchains and getting started guide: remove duplication, other fixes

marc-hb:
Fix the CMake example, it has been broken since defconfig change
https://github.com/thesofproject/sof/pull/3821

Fix the default target when building tools.

Replace a lot of the massive copy/paste with "for platform" loops.

Move PATH changes to newlib section and explain that they don't need to
be permanent.

### DIFF
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -152,9 +152,13 @@ Create or append to the ``LD_LIBRARY_PATH`` environment variable.
 Step 2 Build toolchains from source
 ===================================
 
-Build the xtensa cross-compilation toolchains with crosstool-ng for Intel |BYT|,
-|CHT|, |HSW|, |BDW|, |APL|, |CNL|, |ICL|, |JSL|, |TGL| platforms and NXP i.MX8/i.MX8X/i.MX8M
-platforms.
+Build the xtensa cross-compilation toolchains with crosstool-ng for
+Intel |BYT|, |CHT|, |HSW|, |BDW|, |APL|, |CNL|, |ICL|, |JSL|, |TGL|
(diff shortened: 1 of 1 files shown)


getting_started/build-guide/build-from-scratch.rst
```diff
@@ -221,13 +225,26 @@ default values missing from it:
...
    diff -u config-apl-gcc10.2-gdb9 .config
 
-"Install" toolchains by copying them to ``$SOF_WORKSPACE``.
+While other steps take minutes at most, building all toolchains may last
+about an hour depending on the performance of your system. Run this loop
+to build all toolchains without interruption:
+
+.. code-block:: bash
+
+   time for i in config*gcc10*; do
```
ujfalusi, line 234: we have two gcc10 config for apl:
```
config-apl-gcc10.2-gdb8.3
config-apl-gcc10.2-gdb9
```
Will this going to cause issue?

lgirdwood: lets use gdb9

lgirdwood (commented):
Good to merge once we grep on gdb9

kv2019i (approved):

getting_started/build-guide/build-from-scratch.rst
marc-hb: > we have two gcc10 config for apl:

Great catch, thanks.

> Will this going to cause issue?

I think I got away with it because globbing is "sorted according to the collating sequence in effect in the current locale."  https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_13_03

> let's use gdb9

Done, changed to `config*gcc10.2-gdb9`

lgirdwood (approved):

Issue #347, opened 2021-05-23
End-user documentation and bug reporting


fxthomas:
I'm not sure if this is the correct place for documentation requests, feel free to point me to the right place.

I had a rather difficult time finding SOF firmware logs with the official kernel from my distribution when trying to debug https://github.com/thesofproject/sof/issues/3868. One of the reasons is that it is still not clear to me, as an end-user relatively unfamiliar with SOF development, exactly how to get logs in a reliable way in either my own or my distribution's kernel builds.

The questions below are things I'm wondering or found out while doing research.

----

**What's the difference between `trace` and `etrace`?**

The documentation on the [Logger](https://thesofproject.github.io/latest/developer_guides/debugability/logger/index.html) mentions the existence of two files, `/sys/kernel/debug/sof/trace` and `/sys/kernel/debug/sof/etrace`.

* What are they?
* Why two of them?

Since there are two optinos, and my distro's official kernel only has `trace` enabled by default, my hypothesis is that `etrace` has more overhead/is more detailed, and as such is not enabled in default builds. Am I correct?

**How to enable log tracing?**

By default on my distro's official build `trace` is empty and `etrace` doesn't exist, but in custom builds I do following the [build instructions](https://thesofproject.github.io/latest/getting_started/build-guide/build-from-scratch.html) both show messages.

* What kernel compilation options control them? Should I need only CONFIG_SND_SOC_SOF_DEBUG and CONFIG_SND_SOC_SOF_DEBUG_ENABLE_FIRMWARE_TRACE?
* Can I enable them dynamically without having to recompile my distro's official kernel?

**What is dynamic debug, is that different from the logger?**

[This page](https://thesofproject.github.io/latest/getting_started/intel_debug/suggestions.html#enable-dynamic-debug) suggests to use dynamic debug options. Should I expect to get different kinds of information than from the logger?

paulstelian97:
I'll give my two cents since I know a bit about the logger.

So, `trace` is so-called DMA trace. The `sof-logger` tool will block on it until the firmware gives more logs, then will decode (by using the .ldc file associated with the running firmware, which you must manually provide in the command currently) those logs from the binary format to a user-readable format.

[...]

PR #351, opened 2021-05-31
Update some debugging information

ujfalusi:
Add clarifications and details for debugging SOF issues.

The updates are created based on issue #347 in hope that the information in the documentation will be more user friendly.

### DIFF
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -128,6 +128,14 @@ Examples
      sof-logger -l ldc_file -d
 
 
+.. note::
+  debugfs files used by ``sof-logger``
+
+  - ``etrace``: direct access to the shared TRACE window of the SOF firmware
+  - ``trace``: using DMA to stream debug trace information from SOF firmware (on
+    distribution kernels ``sof_debug=1`` module option for ``snd_sof`` might be
+    needed if the /sys/kernel/debug/sof/trace file is not present)
+
 Trace filtering
 ***************
 
--- a/getting_started/intel_debug/suggestions.rst
+++ b/getting_started/intel_debug/suggestions.rst
@@ -117,6 +117,9 @@ file:
 
 Note that this list is only an example.
 
+Dynamic debug is a Linux kernel feature. For detailed information, see the
+official `kernel documentation <https://www.kernel.org/doc/html/latest/admin-guide/dynamic-debug-howto.html>`__.
+
 Install sof-logger
 ******************
 
--- a/getting_started/intel_debug/suggestions.rst
+++ b/getting_started/intel_debug/suggestions.rst
@@ -131,4 +131,12 @@ following command to extract DSP traces:
 
 .. code-block:: bash
 
-   sof-logger -t sof-tgl.ldc
+   sof-logger -t -l sof-tgl.ldc
+
+Trace support might need to be enabled on distribution kernels in case the
+``/sys/kernel/debug/sof/trace`` file is not present by adding sof_debug=1 option
+to snd_sof module:
+
+.. code-block::
+
+   options snd_sof sof_debug=1


getting_started/intel_debug/suggestions.rst
```diff
@@ -131,4 +131,12 @@ following command to extract DSP traces:
 
 .. code-block:: bash
 
-   sof-logger -t sof-tgl.ldc
\ No newline at end of file
+   sof-logger -t sof-tgl.ldc
```
lyakh, line 134: I think you need a `-l` option before the `.ldc` file?

developer_guides/debugability/logger/index.rst
```diff
@@ -128,6 +128,15 @@ Examples
      sof-logger -l ldc_file -d
 
 
+.. note::
+  debugfs files used by ``sof-logger``
+
+  - ``etrace``: direct access (memcpy) to TRACE window of the SOF firmware
+  - ``trace``: Using DMA to stream debug trace information from SOF firmware (on
```
lyakh, line 135: either both capital or both lowercase "direct" and "using"

lyakh (changes requested):
the missing `-l` was there before, but since you already touch that text, we can fix it as well

developer_guides/debugability/logger/index.rst
ujfalusi: OK, I'll convert to lowercase.

getting_started/intel_debug/suggestions.rst
ujfalusi: I think you are right, I will fix it while at it. I'm not sure why the change is registered, probably because the new lines.

kv2019i: Good that it grabbed the line... this needs to be fixed (although unrelated to your patch).

kv2019i (approved):

lgirdwood (approved):

developer_guides/debugability/logger/index.rst
```diff
@@ -128,6 +128,15 @@ Examples
      sof-logger -l ldc_file -d
 
 
+.. note::
+  debugfs files used by ``sof-logger``
+
+  - ``etrace``: direct access (memcpy) to TRACE window of the SOF firmware
```
marc-hb, line 134:
```suggestion
  - ``etrace``: direct access to TRACE window of the SOF firmware in the shared memory "mailbox"
```

marc-hb (commented):
The term "mailbox" is all over the code and probably in the documentation too so it's useful to mention it here.

developer_guides/debugability/logger/index.rst
ujfalusi: @marc-hb, the trace window is not within the 'mailbox' memory. We have a bigger shared area and within that area we have the 'mailbox' (up and down region), trace, debug, stream, regs and exception windows.
Some of these windows are optional and only used for debugging, like the trace region.

Would it be better if we use the term 'region' instead of 'window' when referring to the TRACE window/region of the shared area?

ujfalusi: However the sof-logger's help documents the t option:
`sof-logger:      -t                     DMA 'trace' stream instead of 'etrace' error mailbox`

Probably something like this:
etrace: direct access to TRACE error mailbox of the firmware

marc-hb: >  We have a bigger shared area and within that area we have the 'mailbox' (up and down region), trace, debug, stream, regs and exception windows.

Thanks, can you please share some pointers to this information? It looks like our terminology is inconsistent.

> However the sof-logger's help documents the t option:

_I_ recently changed this help message so it does not count :-) However I changed it based on symbols like `CONFIG_TRACEM` and `MAILBOX_TRACE_BASE` and other comments `git grep -i -e mailbox -e mbox -- src/trace/`; most of these are not from me.

So maybe "mailbox" is not the correct term here. On the other hand, "shared memory" is IMHO still better than "(memcpy)" in documentation.

ujfalusi: I would go with a simple:
`etrace: direct access to TRACE window of the SOF firmware`
If ever we will support non mem mapped type of DSP with SOF then this will be still valid.
We can use `region` instead of `window`, that should be fine as well.

The regions/windows are defined under xsram_window and they are really just sections within a shared memory between the CPU and DSP in currently supported platforms.
Imho to say that something is a mailbox it would take a bit more than a shared memory, a signaling mechanism like doorbells should be there. TRACE does not have associated doorbell afaik.

deb-intel: @lyakh Have your change requests been completed? @ujfalusi Are you ready for the document to be merged? thanks

marc-hb: > I would go with a simple: etrace: direct access to TRACE window of the SOF firmware

I feel like this is a confusing way to say "non-DMA trace" or "shared memory". What's wrong with either of these?

lyakh (approved):

developer_guides/debugability/logger/index.rst
lyakh: > @lyakh Have your change requests been completed? @ujfalusi Are you ready for the document to be merged? thanks

@deb-intel yes, they have! I've approved the PR

ujfalusi: @marc-hb, can you check the new wording for etrace:
``etrace``: direct access to the shared TRACE window of the SOF firmware

marc-hb (approved):

lgirdwood:
@deb-intel fyi - ready for grammar check :)

deb-intel:
@lrgirdwo Sounds good. I plan to publish the site today.

Issue #352, opened 2021-06-09
Undefined AM_GNU_GETTEXT when building alsa-utils [WSL2 Ubuntu 20.04 LTS (Windows)]


edward-a:
I am following Getting Started instructions to build SOF on WSL2 Ubuntu 20.04 LTS (Windows) and i was able to successfully compile ALSA libraries. I am however stuck on compiling alsa-utils with the following errors:

configure.ac:10: error: possibly undefined macro: AM_GNU_GETTEXT
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:11: error: possibly undefined macro: AM_GNU_GETTEXT_VERSION

Here below is the complete output after executing ./gitcompile

configure.ac:10: warning: macro 'AM_GNU_GETTEXT' not found in library
configure.ac:11: warning: macro 'AM_GNU_GETTEXT_VERSION' not found in library
./gitcompile: line 10: gettextize: command not found
configure.ac:15: installing './compile'
configure.ac:6: installing './install-sh'
configure.ac:6: installing './missing'
alsactl/Makefile.am: installing './depcomp'
axfer/test/Makefile.am:9: warning: source file '../container.c' is in a subdirectory,
axfer/test/Makefile.am:9: but option 'subdir-objects' is disabled
automake: warning: possible forward-incompatibility.
automake: At least a source file is in a subdirectory, but the 'subdir-objects'
automake: automake option hasn't been enabled.  For now, the corresponding output
automake: object file(s) will be placed in the top-level directory.  However,
automake: this behaviour will change in future Automake versions: they will
automake: unconditionally cause object files to be placed in the same subdirectory
automake: of the corresponding sources.
automake: You are advised to start using 'subdir-objects' option throughout your
automake: project, to avoid future incompatibilities.
axfer/test/Makefile.am:9: warning: source file '../container-riff-wave.c' is in a subdirectory,
axfer/test/Makefile.am:9: but option 'subdir-objects' is disabled
axfer/test/Makefile.am:9: warning: source file '../container-au.c' is in a subdirectory,
axfer/test/Makefile.am:9: but option 'subdir-objects' is disabled
axfer/test/Makefile.am:9: warning: source file '../container-voc.c' is in a subdirectory,
axfer/test/Makefile.am:9: but option 'subdir-objects' is disabled
axfer/test/Makefile.am:9: warning: source file '../container-raw.c' is in a subdirectory,
[...]

PR #353, opened 2021-06-18
build-guide: Correct ct-ng path for building all toolchains

ujfalusi:
There is no mention in the documentation about "ct-install", the ct-ng is
installed in previous steps to $(pwd).

Fixes: a31ee37bae5f2 ("Toolchains and getting started guide: remove duplication, other fixes")

### DIFF
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -232,7 +232,7 @@ to build all toolchains without interruption:
 .. code-block:: bash
 
    time for i in config*gcc10.2-gdb9; do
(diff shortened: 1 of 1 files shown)


ujfalusi:
This issue got noticed in #352 , fix it right away to make sure that new developers will not hit the same problem.

marc-hb (approved):
oops, my bad: I used that different `--prefix` locally and leaked it here. Apologies.

PS: `--prefix=$(pwd)` seems wrong but that's for another day.

ujfalusi:
@marc-hb, I have missed it in the review as well...

lgirdwood (approved):

PR #357, opened 2021-06-24
build-guide: Updates based on a clean Ubuntu 20.04 installation

ujfalusi:
I have taken the time to install Ubuntu 20.04.2 to VBox.

Based on that I would suggest the following clarifications and fixes to the
build guide.

### DIFF
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -16,7 +16,7 @@ Build SOF
 *********
 
 The following steps describe how to install the SOF development
-environment on Ubuntu 16.04, 18.04, and 18.10. They should work on
+environment on Ubuntu 16.04, 18.04, 20.04, and 18.10. They should work on
 19.04, 19.10 and other Linux distributions with minor or no
 modifications.
 
@@ -30,13 +30,16 @@ modifications.
    because some default configuration files refer to other clones using
    relative locations like ``../sof/``.
 
+   Make sure that ``$SOF_WORKSPACE`` have adequate disk space when building the
+   toolchain. About 15GB is going to be needed per toolchain.
+
 Step 0 Set up the workspace directory
 =====================================
 
   .. code-block:: bash
 
      SOF_WORKSPACE=~/work/sof
-     mkdir "$SOF_WORKSPACE"
+     mkdir -p "$SOF_WORKSPACE"
 
 Step 1 Set up build environment
 ===============================
@@ -58,6 +61,13 @@ Make sure that ``build-essential`` and ``git`` are installed:
 .. code-block:: bash
 
    sudo apt-get install build-essential git
+
+* For Ubuntu 20.04:
+
+  .. code-block:: bash
+
+     sudo apt install autoconf flex bison texinfo help2man gawk libtool-bin \
+	libncurses5 libncurses5-dev libssl-dev libgtk-3-dev tree ninja
 
 * For Ubuntu 18.10:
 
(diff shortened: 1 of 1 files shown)


ujfalusi:
The disk space note is added because I first created the VM with click-click-click and the default is 10G.
I was not able to build the first toolchain do to out of space and it took a while to figure out how to re-size the virtual disk then the partition to resume...

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -30,13 +30,16 @@ modifications.
    because some default configuration files refer to other clones using
    relative locations like ``../sof/``.
 
+   Make sure that ``$SOF_WORKSPACE`` have adequate amount of disk space. 100GB
+   or more is recommended.
```
marc-hb, line 34: Please say most of it is for the toolchains because many people don't bother creating the toolchains and use the docker image instead.

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -261,6 +271,12 @@ from ``$SOF_WORKSPACE`` to them:
    cd "$SOF_WORKSPACE"
    for i in crosstool-ng/builds/xtensa-*; do ln -s "$i"; done
 
+Remove the temporarily build files (~7GB per toolchain)
+
+.. code-block:: bash
+
+   rm -rf .build
```
marc-hb, line 278: Wow, great catch! Impressive. Why hide 7G in a hidden directory? Makes really no sense to me. gitignore is enough (I even tend to remove build directories from gitgnore locally. I digress)

marc-hb (commented):
One small but IMHO important suggestion below.

Thanks for taking the time to test all this. I'm surprised you found so few issues! We were not that bad after all.

getting_started/build-guide/build-from-scratch.rst
lgirdwood: So this is only if we build the toolchain, we could have that as an option, probably making the download as the default .

ujfalusi: > Please say most of it is for the toolchains because many people don't bother creating the toolchains and use the docker image instead.

@marc-hb, I'll update to be specific that the space is needed for building the toolchain. This is the build from scratch guide, not the build with docker.

dbaluta: @ujfalusi I wonder if we can make the toolchains available somewhere so the user only needs to download them.

marc-hb: > This is the build from scratch guide, not the build with docker.

Sorry I forgot that "build from scratch" means "not with Docker". About half of the "build from scratch" content applies to Docker the same, now I realize that there has been some serious copy/paste/diverge between the two. I suggest maintaining that "common" half like it would actually become common in the future.

>  I wonder if we can make the toolchains available somewhere so the user only needs to download them.

That would be nice. This by far the most disk, CPU and time-consuming part of this section.

ujfalusi: @dbaluta, I don't know if we have the infra for file server, but it would be a good thing.
Personally I feel safer to build it for myself, it has to be done only once and I know that it will work on my non Ubuntu Linux ;)

@marc-hb, @lrgirdwo, do you have suggestions on storing the toolchains?

marc-hb: > @marc-hb, @lrgirdwo, do you have suggestions on storing the toolchains?

The most natural place would be either at https://github.com/thesofproject/xtensa-overlay/releases or at https://github.com/thesofproject/crosstool-ng/releases and - surprise - there is already something in both places.

ujfalusi: @marc-hb, great. I can send an update later to document the option to download and not build the toolchain form the releases of both xtensa and crosstool-ng.
Most likely later this week or next week?

dbaluta: @marc-hb cool!!

lgirdwood (approved):

PR #373, opened 2021-10-05
Add cmake and ninja-build to the installation instructions

anton-intel:
cmake and ninja-build are required to run documentation processors, but these tools were not mentioned in the "Install documentation tools" section.

### DIFF
--- a/contribute/process/docbuild.rst
+++ b/contribute/process/docbuild.rst
@@ -131,14 +131,14 @@ Depending on your Linux version, install the following tools:
   .. code-block:: bash
 
      sudo apt-get install doxygen python3-pip python3-wheel make \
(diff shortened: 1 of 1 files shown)


lgirdwood (commented):
@anton-intel can you say in the commit message why we need to add cmake and ninja-build. Thanks
@marc-hb fyi

marc-hb (approved):
I guess sof-docs does not require cmake and ninja itself strictly speaking but we document how to also run sof/doxygen and that requires cmake.

We could probably make ninja optional but I don't see any value in that. It's really become standard now.

deb-intel:
@lgirdwood, given @marc-hb's comment, I'm going to merge this PR.

PR #376, opened 2021-10-11
Edit the topic Build SOF from scratch

anton-intel:
1)	Updated dependencies for ubuntu 20.04
2)	Added instructions on how to install alsa-lib and alsa-utils locally
3)	Changed header levels to show them in TOC
4)	Minor edits

### DIFF
--- a/getting_started/build-guide/build-from-scratch.rst
+++ b/getting_started/build-guide/build-from-scratch.rst
@@ -1,7 +1,7 @@
 .. _build-from-scratch:
 
-Build SOF from scratch
-######################
+Build toolchains and SOF from sources
+#####################################
 
 .. contents::
    :local:
@@ -19,9 +19,9 @@ modifications.
 
 .. note::
 
-   Building |SOF| from scratch might take several hours. We recommend
-   that you use Docker to build SOF. For more information, see
-   :ref:`build-with-docker`.
+   Building the toolchains from source might take several hours. We
+   recommend that you use Docker to build SOF. For more information,
+   see :ref:`build-with-docker`.
 
 Step 1. Set up the workspace directory
 **************************************
@@ -34,8 +34,9 @@ directory. Clone all git repositories at the same directory level
 because some default configuration files refer to other clones using
 relative locations like ``../sof/``.
 
-Make sure that ``$SOF_WORKSPACE`` has adequate disk space when building
-the toolchain. About 15GB is needed per toolchain.
+Make sure that ``$SOF_WORKSPACE`` has adequate disk space when
+building the toolchain. About 15GB is needed per toolchain. You can
+reclaim some of the disk space after building the toolchain.
 
   .. code-block:: bash
 
@@ -113,8 +114,8 @@ If you use Ubuntu 18.04+ you can install CMake with apt:
 For Ubuntu 16.04, CMake from apt is outdated and you must install CMake from
 sources. Refer to this short guide: https://cmake.org/install/.
 
-Build alsa-lib and alsa-utils
-=============================
+Build alsa-lib and alsa-utils from source
+=========================================
 
 This project requires some new features in :git-alsa:`alsa-lib` and
 :git-alsa:`alsa-utils`, so build the newest ALSA from source code.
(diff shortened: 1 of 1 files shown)


dbaluta (approved):

paulstelian97 (approved):

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -12,40 +12,41 @@ Intel platforms include: |BYT|, |CHT|, |HSW|, |BDW|, |APL|, |CNL|, |ICL|, |JSL|,
...
 19.04, 19.10 and other Linux distributions with minor or no
 modifications.
 
 .. note::
 
-   Point the ``$SOF_WORKSPACE`` environment variable to the directory in
-   which you store all sof work.
+   Building |SOF| from scratch might take several hours. We recommend
+   that you use Docker to build SOF. For more information, see
+   :ref:`build-with-docker`.
```
marc-hb, line 24: That's oversimplified and too negative, gives a bad first impression. It sounds like a one-line change in a some `sof/common.h` file might take several hours where in reality that costs less than a minute to rebuild all platforms. There is only a single step that takes hours and that's building the toolchains and it's required only once and then never again.

The problem is the name of this section. It's technically not "Build SOF from scratch" but "Build everything from scratch". Can you change that section name?

Then the start of this paragraph can be changed to "Building the toolchains from scratch might take..."

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -365,9 +378,21 @@ can skip the next two sections.
 The installer also builds and deploys some user-space binaries from the
 ``sof/tools/`` subdirectory.
 
+.. note::
+
+   If interrupted, the installer process might leave corrupted files
+   in the ``installer-builds`` subdirectory. In this case, all
+   subsequent attempts to run the installer end with an error. To
+   resolve this issue, delete the ``installer-builds`` directory and
+   run the installer again.
```
marc-hb, line 387: By chance I got for the first time ever in the same error message you reported to me by email. It didn't happen because the build was interrupted but because I tried to switch toolchains without deleting the build directory. This should be very rare and all developers' first reflex when they don't understand some build error is to delete the build directorie(s).

Did you ever experience that same error again?

In any case the installer is merely invoking the other build scripts, so if you really want to add such a warning it should not sound like the installer builds in a unique and different way. How about something like this:

> The installer is much faster than the lower level `./scripts/` it relies on because it does not start by deleting the build directories every time it is invoked.  However some "big" configuration changes like switching to a different toolchain or some rare build failures can leave the `installer-builds/build_*` directories in an inappropriate state. In such a case just delete them and run the installer again.

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -558,18 +585,20 @@ the kconfig repo, and the :ref:`sof_driver_arch`.
       make defconfig
       git clone https://github.com/thesofproject/kconfig
       scripts/kconfig/merge_config.sh .config ./kconfig/base-defconfig ./kconfig/sof-defconfig  ./kconfig/mach-driver-defconfig ./kconfig/hdaudio-codecs-defconfig
-      (optional) make menuconfig
 
-   Select the SOF driver support and disable SST drivers.
+   Optionally, you can also run ``make menuconfig``, navigate to
+   Device Drivers > Sound card support > Advanced Linux Sound
+   Architecture, and select the **Prefer SOF driver over SST on BY/CHT
+   platforms** option.
```
marc-hb, line 592: cc: @plbossart

getting_started/build-guide/build-from-scratch.rst
```diff
@@ -12,40 +12,41 @@ Intel platforms include: |BYT|, |CHT|, |HSW|, |BDW|, |APL|, |CNL|, |ICL|, |JSL|,
...
 
-Step 0 Set up the workspace directory
-=====================================
+The code examples assume ``$SOF_WORKSPACE`` is the top-level working
+directory. Clone all git repositories at the same directory level
+because some default configuration files refer to other clones using
+relative locations like ``../sof/``.
+
+Make sure that ``$SOF_WORKSPACE`` has adequate disk space when building
+the toolchain. About 15GB is needed per toolchain.
```
marc-hb, line 38: I think it's possible to reclaim a lot of that once it's done, @ujfalusi ?

getting_started/build-guide/build-from-scratch.rst
ujfalusi: Or a better section name would be to 'build the development environment from sources', ''set up the development environment from sources' or along these lines.

PR #381, opened 2021-11-10
debugability: logger: add note on DMA trace limitations

kv2019i:
If sof-logger is started (or restarted) while DSP is running,
the initial traces may come in incorrect order or are incomplete.
This is important to note when parsing the logger results.

BugLink: https://github.com/thesofproject/sof-test/issues/297

### DIFF
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -136,6 +136,20 @@ Examples
     distribution kernels ``sof_debug=1`` module option for ``snd_sof`` might be
     needed if the /sys/kernel/debug/sof/trace file is not present)
 
+.. note::
+  If sof-logger is started (or restarted) while firmware is active, the initial
+  trace messages might be incomplete. This can happen as current (as of SOF version
+  1.9) trace is implemented with a ringbuffer between firmware and the kernel driver.
+  When sof-logger is started, kernel will always start to read the ringbuffer from
+  0 position, independently of firmware state. If firmware write pointer just wrapped
+  around when sof-logger is started, sof-logger will only show the traces after
+  the wrap. Firmware write position is also reset whenever firmware is booted,
+  including runtime suspend and resume. To capture traces over runtime suspend
+  events, the kernel trace interface will signal end of file at runtime suspend.
+  When sof-logger notices the end of file marker, it will reopen the trace
+  file and start reading from position 0 and thus be in sync with the firmware
+  when it is resumed.
+
 Trace filtering
 ***************
 


paulstelian97 (approved):

developer_guides/debugability/logger/index.rst
```diff
@@ -136,6 +136,11 @@ Examples
     distribution kernels ``sof_debug=1`` module option for ``snd_sof`` might be
     needed if the /sys/kernel/debug/sof/trace file is not present)
 
+.. note::
+  The DMA trace sysfs interface is a streaming interface. If sof-logger is
+  started (or restarted) while DSP is active, it is possible the initial traces
```
ujfalusi, line 141: typo: s/traces/trace

ujfalusi (changes requested):
One typo in the text.

I'm not sure if it can be put into words, but over time the trace messages (as they flow in) will start to straightened up. Iow, the first batch of trace information might not be of an relevance of the past. Or something.

developer_guides/debugability/logger/index.rst
anton-intel: Would the following work?

The DMA trace sysfs interface is a streaming interface. If sof-logger is started (or restarted) while DSP is active, the initial trace messages might appear in the wrong order or incomplete.

marc-hb (commented):
>  If sof-logger is  started (or restarted) while DSP is active, ...

That does not really make sense because the logger can never run unless the driver and firmware have been loaded. When you try:

```
sudo sof-logger  -t -f 3 -l  /lib/firmware/intel/sof/sof-apl.ldc 
error: Unable to open version file /sys/kernel/debug/sof/fw_version: No such file or directory
```

The description also does not match the following behavior: after disabling D3 with `echo on > sys/devices/pci0000:00/"${pci_id}"/power/control` , the sof-logger can be stopped and restarted any number of times and it displays the exact same DMA traces every single time. That could not possibly happen if it were really implemented as a streaming interface.

Even worse: multiple instances of the sof-logger show the same DMA traces even when run concurrently.

So this description does not seem to match reality.

ujfalusi:
> > If sof-logger is  started (or restarted) while DSP is active, ...
> 
> That does not really make sense because the logger can never run unless the driver and firmware have been loaded. When you try:
> 
> ```
> sudo sof-logger  -t -f 3 -l  /lib/firmware/intel/sof/sof-apl.ldc 
> error: Unable to open version file /sys/kernel/debug/sof/fw_version: No such file or directory
> ```

@marc-hb, do we really need to document that if you want to collect dtrace from SOFirmware  you need to have the drivers loaded?
If the driver probe fails then the drivers are not loaded and the debugfs is not populated, so there will be no dtrace, fw_version or even sof directory.
It is another question if this is the right thing to do... What I mean is that it might be helpful to have at least access to the mtrace logs to be able analyze the failure? Hrm, we might want to have a flag to sof_debug to instruct the driver to dump the mtrace in case of boot fail?

> The description also does not match the following behavior: after disabling D3 with `echo on > sys/devices/pci0000:00/"${pci_id}"/power/control` , the sof-logger can be stopped and restarted any number of times and it displays the exact same DMA traces every single time. That could not possibly happen if it were really implemented as a streaming interface.
> 
> Even worse: multiple instances of the sof-logger show the same DMA traces even when run concurrently.
> 
> So this description does not seem to match reality.

Yes, I think that is the behavior atm or something similar.
If I look at the dtrace code and logic on the firmware side and on the kernel side I think (again, imho) we have some misaligned assumptions and bugs on both sides which allows it to work but in a non defined way (well,if you read the code you can see the 'order' in the chaos).
The kernel driver uses the position within the debugfs file stream as read offset on the dtrace buffer along with a position information coming from the firmware. They are unrelated. When you start sof-logger the position in the file is 0, then you follow the trace and it is fine, but if you stop it and restart it: the position in the trace file is going to start from 0 and it does not matter if we killed the sof-logger when it was in mid buffer (offset derived from the position).
If you start/stop/start/stop/start/stop sof-logger it will always going to read from 0 offset of the buffer.

This is only one - and obvious - bug, but if you look closer there are other cases where things get out of control.

I think we should fix the dtrace for good and then update the documentation or even better have a description what we want.
I know that if I fix the dtrace then there will be complaints that it did things differently in the past (no matter if it was wrong)...

marc-hb:
> @marc-hb, do we really need to document that if you want to collect dtrace from SOFirmware you need to have the drivers loaded?

No, I was merely describing one of the mismatches between this PR and reality. That's all.

> I know that if I fix the dtrace then there will be complaints that it did things differently in the past (no matter if it was wrong)...

I doubt that. I think the sof-logger expectations are pretty low so I am optimistic that the only "complaints" will be from sof-test - I mean test failures. I will align sof-test to all "correct" and well defined behaviors that match (future) reality.

So if you know what to fix then please go for it.

ujfalusi:
> > I know that if I fix the dtrace then there will be complaints that it did things differently in the past (no matter if it was wrong)...
> 
> I doubt that. I think the sof-logger expectations are pretty low so I am optimistic that the only "complaints" will be from sof-test - I mean test failures. I will align sof-test to all "correct" and well defined behaviors that match (future) reality.

Right. One thing which bugs me is around the `sof-logger` start/stop/start/stop.
We either do it in a simple manner: `sof-logger` will only show the trace from the point it started.
or
We try to do a more sophisticated (and might better) approach by fetching logs not have been taken by a previous run of `sof-logger`. We need to deal with ring buffer wrapping and stuff to get chronological backlog and it would be nice to inject (in kernel level) some marks as well.

> So if you know what to fix then please go for it.

I kind of know what to fix, I'm debating the _how_ ;)

marc-hb:
> We try to do a more sophisticated (and might be better) approach by fetching logs not have been taken by a previous run of sof-logger

Not losing logs: that's the least I would expect from a "streaming" interface like what this PR describes (ignoring ringbuffer overruns of course).

ujfalusi:
> > We try to do a more sophisticated (and might be better) approach by fetching logs not have been taken by a previous run of sof-logger
> 
> Not losing logs: that's the least I would expect from a "streaming" interface like what this PR describes (ignoring ringbuffer overruns of course).

It depends if it supports time shifting ;) "streaming" is real time, not on-demand, but I also prefer to have a bit of context from the past (from the point the previous sof-logger was stopped - if it run).

kv2019i:
@marc-hb I'll remove "streaming", you are reading too much into it.

kv2019i:
Updated. Thanks @anton-intel  , @ujfalusi  and @marc-hb for the comments.

ujfalusi (approved):

kv2019i:
@marc-hb wrote:
> The description also does not match the following behavior: after disabling D3 with echo on 
> sys/devices/pci0000:00/"${pci_id}"/power/control , the sof-logger can be stopped and restarted any number of times and it 
> displays the exact same DMA traces every single time. That could not possibly happen if it were really implemented as a 
> streaming interface.

I removed "streaming" now. Now behavior matches description. ""If sof-logger is started (or restarted) while DSP is active, the initial trace messages **might appear in the wrong order or be incomplete**.""

Don't shoot the messenger here, that's just how it works. The fact that you get traces most of the time upon restart, just makes this all the more dangerous (and worthy of note) as IT IS NOT GUARANTEED!

anton-intel (approved):

kv2019i:
@lgirdwood Let me remeasure the DSP load with the mixer+eq optimizations now merged. I'll then make a proposal wether the mixer should be dropped or not. I'll keep this as draft until that point.

marc-hb:
> Don't shoot the messenger here, that's just how it works. The fact that you get traces most of the time upon restart, just makes this all the more dangerous (and worthy of note) as IT IS NOT GUARANTEED!

Sorry for shooting the messenger but when a NOT GUARANTEED behavior works 99.9% of the time then something is very deeply flawed and the documentation should not look like everything is fine.

marc-hb:
> but I also prefer to have a bit of context from the past (from the point the previous sof-logger was stopped - if it run).

Right, otherwise you can't debug... boot!

marc-hb:
As stated in various other github places, I think this is too short. It should give at least clues why using keywords like "ringbuffer", "offset" and "DSP suspend".

Maybe also worth noting that we do this in sof-test for every test.

kv2019i:
@marc-hb wrote:
> Sorry for shooting the messenger but when a NOT GUARANTEED behavior like this one works 99.9% of the time then something is very deeply flawed and the documentation should not look like everything is fine.

I don't really agree with the general statement. Many systems have timing uncertainty and a compromise needs to be made between 100%, 99.9% and 90% odds (with required work effort to reach a level of certainly varying accordingly).

I expanded the explanation of current behaviour and repushed the PR. 

I also filed an enhancement ticket to the kernel interface: https://github.com/thesofproject/linux/issues/3275
If given the current design and its limitations, we cannot reliably implement automated validation of FW traces, then the design is not fit for purpose and we need to change it.

lgirdwood (approved):
Thanks @kv2019i, this is useful information whilst we all align on a fix.

marc-hb:
thanks @kv2019i for the extended version, I find it extremely useful.

PR #388, opened 2021-11-23
Correction: Build SOF with Docker

anton-intel:
Make the user choice between the Docker image build options clearer.


marc-hb (approved):
At a higher level, I wonder how clear it is that Docker is optional as a whole.

lgirdwood (approved):

lgirdwood:
> At a higher level, I wonder how clear it is that Docker is optional as a whole.

Good point. @anton-intel are you able to make it clear docker is optional. i.e. it can be used where developers dont want to build toolchains and dont want to update system ALSA libraries/tools.

PR #391, opened 2021-12-09
developer_guides: add coredump-reader example usage on Linux

kv2019i:
Add documentation and practical example how to extract the DSP
coredump via Linux SOF driver debugfs interface and how to
parse the result with coredump-reader.

### DIFF
--- a/developer_guides/debugability/coredump-reader/index.rst
+++ b/developer_guides/debugability/coredump-reader/index.rst
@@ -30,3 +30,53 @@ We read from dump file into sof-coredump-reader.py, then we pipe its output to x
 .. code-block:: bash
 
    ./sof-coredump-to-gdb.sh sof-apl dump_file
+
+Usage with Linux SOF Driver
+***************************
+
+If a core dump occurs after a DSP error, the Linux SOF driver allows
+accessing the dump via debugfs. Consider the following example of capturing
+the dump file and processing it with coredump-reader:
+
+.. code-block:: bash
+
+   dut> cat /sys/kernel/debug/sof/exception >dsp-coredump
+   # transfer file to host
+   host> sof/tools/coredumper/sof-coredump-reader.py  -v -l 4 -i dsp-coredump -o dsp-coredump.gdb
+   host> xt-gdb sof/build_tlg_xcc/sof --command=dsp-coredump.gdb
+   [cut]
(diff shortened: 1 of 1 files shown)


developer_guides/debugability/coredump-reader/index.rst
```diff
@@ -30,3 +30,47 @@ We read from dump file into sof-coredump-reader.py, then we pipe its output to x
...
+   #5  0xbe02fb29 in ipc_glb_debug_message (header=394016) at /home/user/sof/src/ipc/handler-ipc3.c:1373
+   [cut]
+   (xt-gdb) info all-registers
+   pc             0xbe051b00       0xbe051b00 <literals>
+   ar0            0x0      0
+   ar1            0xbe00a044       -1107255228
+   ar2            0x10000  65536
+
+Notes:
+ - coredump reader only works with xcc toolchain
```
cujomalainey, line 66: Is this simply because `xt-gdb` is used? Or is there another issue if gcc or clang is used?

cujomalainey (commented):
Thanks for the docs, much appreciated :)

developer_guides/debugability/coredump-reader/index.rst
kv2019i: @cujomalainey It's primarily due to xt-gdb. That should be "just work" to fix in the coredumper script, but we've also had some weird issues with gcc built binaries. See e.g. this old bug https://github.com/thesofproject/sof/issues/1346 .

lgirdwood: yeah, so gcc has a slightly different calling ABI AFAICT vs XCC and the stack trace will look wrong. Regular gdb also interprets this incorrectly (but otherwise works at runtime debug on qemu). Maybe something for Maciej to look at in the future.

kv2019i:
Tested doc generation and updated text based on Anton's feedback. Now ready for review.

anton-intel (approved):

greg-intel (approved):
Usage doc looks clean.  Normally I consider collapsible regions on examples and in-depth usage, but in this case it's pretty deep in the folder structure.  So leveraging `<details>` here doesn't make sense.

lgirdwood (approved):

PR #405, opened 2022-02-16
Supported Platforms: Platform Clock column added

mwasko:
Platform clock frequency information is required to get correct timestamps in sof-logger.

### DIFF
--- a/platforms/index.rst
+++ b/platforms/index.rst
@@ -9,27 +9,27 @@ Supported Platforms
 Platform and board specific support is continually added to the SOF project as documented below.
 
 .. csv-table:: Supported Platforms
-   :header: "Platform", "Architecture", "Cores/Clocks", "Memory", "Audio Interfaces"
-   :widths: 20, 20, 10, 10, 20
+   :header: "Platform", "Architecture", "Cores/Clocks", "Platform Clock", "Memory", "Audio Interfaces"
+   :widths: 20, 20, 10, 10, 10, 20
 
-   "Host Testbench", "PC command line", "N/A", "N/A", "N/A Files are used to simulate audio interfaces"
-   "Qemu", "All supported SOF HW platforms", "N/A", "N/A", "WiP Files will be used to simulate audio interfaces"
-   "Intel Baytrail / Merrifield", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "96KB IRAM / 192KB DRAM", "3 x SSP (I2S, PCM)"
-   "Intel Cherrytrail / Braswell", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "96KB IRAM / 192KB DRAM", "6 x SSP (I2S, PCM)"
-   "Intel Broadwell", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "320KB IRAM / 640KB DRAM", "2 x SSP (I2S, PCM)"
-   "Intel Apollolake / Geminilake", "Xtensa HiFi3", "2 @ 100 - 400MHz", "128KB LP SRAM / 512KB HP SRAM", "6 x SSP (I2S, PCM), HDA, DMIC"
-   "Intel Cannonlake / Whiskeylake / Cometlake", "Xtensa HiFi3", "4 @ 120 - 400MHz", "64KB LP / 3008KB HP SRAM", "3 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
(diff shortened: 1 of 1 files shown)


mwasko:
@dbaluta I encourage you to add NXP platforms clock information as well. I have currently left it as TBD.

platforms/index.rst
```diff
@@ -9,27 +9,27 @@ Supported Platforms
 Platform and board specific support is continually added to the SOF project as documented below.
```
dbaluta, line 10: @mwasko what exactly is platform clock? Is the CPU freq?

lgirdwood: Its the timestamp we use for the trace. The userspace tools assume it's 19.2M (unless specified on the cmd line) since this was the time on the initial SOF platforms. It was causing a lot of confusion recently on some platforms that used a different clock...

platforms/index.rst
```diff
@@ -9,27 +9,27 @@ Supported Platforms
...
+   :header: "Platform", "Architecture", "Cores/Clocks", "Platform Clock", "Memory", "Audio Interfaces"
+   :widths: 20, 20, 10, 10, 10, 20
+
+   "Host Testbench", "PC command line", "N/A", "N/A", "N/A", "N/A Files are used to simulate audio interfaces"
+   "Qemu", "All supported SOF HW platforms", "N/A", "N/A", "N/A", "WiP Files will be used to simulate audio interfaces"
+   "Intel Baytrail / Merrifield", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "25MHz", "96KB IRAM / 192KB DRAM", "3 x SSP (I2S, PCM)"
+   "Intel Cherrytrail / Braswell", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "19.2MHz", "96KB IRAM / 192KB DRAM", "6 x SSP (I2S, PCM)"
+   "Intel Broadwell", "Xtensa HiFi2 EP", "1 @ 50 - 400MHz", "24MHz", "320KB IRAM / 640KB DRAM", "2 x SSP (I2S, PCM)"
+   "Intel Apollolake / Geminilake", "Xtensa HiFi3", "2 @ 100 - 400MHz", "19.2MHz", "128KB LP SRAM / 512KB HP SRAM", "6 x SSP (I2S, PCM), HDA, DMIC"
+   "Intel Cannonlake / Whiskeylake / Cometlake", "Xtensa HiFi3", "4 @ 120 - 400MHz", "19.2Hz or 24MHz strap selectable", "64KB LP / 3008KB HP SRAM", "3 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
```
plbossart, line 21: that's not correct @mwasko, we have ZERO support for 19.2MHz in CNL/CML

platforms/index.rst
```diff
@@ -9,27 +9,27 @@ Supported Platforms
...
+   "Intel Suecreek", "Xtensa HiFi3", "2 @ 120 - 400MHz","19.2Hz or 24MHz strap selectable", "64KB LP SRAM / 4096KB HP SRAM", "6 x SSP (I2S, PCM), DMIC"
+   "Intel Icelake", "Xtensa HiFi3", "4 @ 120 - 400MHz", "24MHz or 38.4MHz strap selectable", "64KB LP SRAM / 3008KB HP SRAM", "6 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
+   "Intel Jasperlake", "Xtensa HiFi3", "2 @ 120 - 400MHz", "24MHz or 38.4MHz strap selectable", "64KB LP SRAM / 1024KB HP SRAM", "3 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
+   "Intel Tigerlake", "Xtensa HiFi3", "4 @ 120 - 400MHz", "24MHz or 38.4MHz strap selectable", "64KB LP SRAM / 2944KB HP SRAM", "6 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
+   "Intel Alderlake", "Xtensa HiFi3", "4 @ 120 - 400MHz", "24MHz or 38.4MHz strap selectable", "64KB LP SRAM / 2944KB HP SRAM", "6 x SSP (I2S, PCM), HDA, DMIC, Soundwire"
```
plbossart, line 26: no again, for all of those we ONLY support 38.4.

mwasko: @plbossart if there are no configuration support for 24 MHz then I am okay to limit info to 38.4MHz, the provided data comes from HW spec.

plbossart: right, the problem is what we mean by 'supported'. The SOC can probably support both modes, but in practice we have no platforms with 24 MHz after CNL.

In SOF we have hard-coded assumptions in topologies on clocks, and we also don't have a means to provide the information to the firmware at run time. We discussed about this a while ago in the context of mix-match CPU/chipset permutations and didn't add that mechanism in the end.

PR #412, opened 2022-06-08
build-with-zephyr: fix submodules, get ready for new python script

marc-hb:
Add missing `git submodule update` as reported by @plbossart (thx!)

Also make the documentation ready for `xtensa-build-zephyr.py`
which... does not have any hack to automatically download git submodules
like the older .sh script has.

As a side-effect, this breaks down the process and adds an intermediate
`west build` step which removes one layer of indirection in case of some
Zephyr environment issue.

Cannot fully switch to the Python version yet because of bugs like
https://github.com/thesofproject/sof/pull/5906 but do recommend users to
start trying it.

### DIFF
--- a/getting_started/build-guide/build-with-zephyr.rst
+++ b/getting_started/build-guide/build-with-zephyr.rst
@@ -30,8 +30,8 @@ Check out and build
    .. note::
 
       If you need a different SOF version than the one that west
-      automatically checks out, change to ``modules/audio/sof`` and use git
-      to select your preferred version. You need at least version 1.6 to use
+      automatically checks out, instructions below show how to change ``modules/audio/sof``
+      and select your preferred version. You need at least Zephyr version 1.6 to use
       it with Zephyr. Make sure you branch or tag your code in git;
       otherwise, a future ``west update`` may lose it. See the west user
       guide.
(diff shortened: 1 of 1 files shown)


marc-hb:
The many `breathe` warnings are unrelated to this. They're caused by the recent addition of `__sparse_*` attributes in the source code. 
https://github.com/thesofproject/sof-docs/runs/6803642273?check_suite_focus=true
cc: @lyakh 

```
/home/runner/work/sof-docs/sof-docs/api/audio-stream-api.rst:8: WARNING: Error when parsing function declaration.
If the function has no return type:
  Error in declarator or parameters and qualifiers
  Invalid definition: Expected identifier in nested name, got keyword: int [error at 10]
    static int audio_stream_set_params(struct  audio_stream   __sparse_cache  * buffer, struct  sof_ipc_stream_params  * params)
    ----------^
If the function has a return type:
  Error in declarator or parameters and qualifiers
  If pointer to member declarator:
    Invalid definition: Expected '::' in pointer to member (function). [error at 34]
      static int audio_stream_set_params(struct  audio_stream   __sparse_cache  * buffer, struct  sof_ipc_stream_params  * params)
      ----------------------------------^
  If declarator-id:
    Invalid definition: Expecting "," or ")" in parameters_and_qualifiers, got "*". [error at 74]
      static int audio_stream_set_params(struct  audio_stream   __sparse_cache  * buffer, struct  sof_ipc_stream_params  * params)
      --------------------------------------------------------------------------^
```

marc-hb:
As usual to see a rendered version of the modified file go to "Files changed" -> "..." -> "View File"

plbossart (approved):
thanks @marc-hb

lgirdwood (approved):

PR #415, opened 2022-06-09
getting_started: setup_linux: intall_locally: Add link to Arch Linux …

ujfalusi:
…guide

Arch users tend to bump into new issues due to the fact that they are using
more recent kernels compared to non rolling distro users.

Add a link to Arch's manual kernel compilation and installation page as
well.

### DIFF
--- a/getting_started/setup_linux/install_locally.rst
+++ b/getting_started/setup_linux/install_locally.rst
@@ -10,7 +10,7 @@ Install the Kernel Locally
 Introduction
 ************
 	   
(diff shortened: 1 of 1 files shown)


plbossart (approved):
Thanks @ujfalusi 

"
Arch users tend to bump into new issues due to the fact that they are using
more recent kernels compared to non rolling distro users. 
"

->> But they also seem to have tons of wikis to work-around issues that are not modified even when the issues are fixed :-)

lgirdwood (approved):

PR #417, opened 2022-07-01
Algorithms: Add equalizers design tutorial

singalsu:
This patch adds a section about measuring acoustical speaker
frequency response and design of IIR and FIR equalizers to
improve it.

Signed-off-by: Seppo Ingalsuo <seppo.ingalsuo@linux.intel.com>


ShriramShastry:
I'm not sure how the graphs are constructed, even though the target filter response charts are produced; therefore, a brief explanation of the input type and filter type used would be useful.

Thanks you !!

singalsu:
@lgirdwood Could someone review this please?

singalsu:
> I'm not sure how the graphs are constructed, even though the target filter response charts are produced; therefore, a brief explanation of the input type and filter type used would be useful.
> 
> Thank you !!

Did the text leave to you opens? I know it's harder to review markup language with pictures separately, but the text describes the steps done where the plots were created. Executing with some own tuning case the commands should help explain what happens.

ShriramShastry (approved):
Thank you!!
I'm okay with plot responses.
Ideally, I would conduct my own test and analyse the results.

btian1:
these are great pictures in frequency domain, and very helpful for audio algorithm developers. 
Can we follow up with some pictures in time domain with sweep signal, it is more straightforward for non-audio algorithm developers, it will directly show which part of freq was impacted in a sweep signal.

btian1 (approved):

algos/eq/equalizers_tuning.rst
```diff
@@ -0,0 +1,924 @@
...
+		card 2: Audio [ThinkPad Dock USB Audio], device 0: USB Audio [USB Audio]
+		  Subdevices: 1/1
+		  Subdevice #0: subdevice #0
+		card 3: UMM6 [UMM-6], device 0: USB Audio [USB Audio]
+		  Subdevices: 1/1
+		  Subdevice #0: subdevice #0
+
+The settings file
+
+.. code-block:: bash
```
mmaka1, line 136: When the "bash" is used as the language mode for styling while it is just 'cat' followed by a content of a file that uses different language, some strange color styling appears. Might be better to just paste the content and put the file name next it it.

algos/eq/equalizers_tuning.rst
```diff
@@ -0,0 +1,924 @@
...
+to equalize highly uncertain details of frequency response.
+
+The smoothed version of the response becomes very flat in the
+equalized version. However the simulated raw response still contains a
+lot of ripple especially at high bands. It’s an industry standard to
+use ⅓ octaves smoothing since it quite well matches human ear
+psycho-acoustics. Therefore ⅓ octaves should be the smallest feasible
+width of octaves smoothing to use.
+
+.. figure:: Picture_response_with_smoothing.png
```
mmaka1, line 356: It is recommended to use SVG images because 1) they are scalable without loosing a quality and 2) text labels etc. that are part of the graphics are searchable.

mmaka1 (approved):

algos/eq/equalizers_tuning.rst
singalsu: Yep, octave seems to now have svg as option for print. Hopefully also web browsers are OK with it. I will test it for next contribution.

algos/eq/equalizers_tuning.rst
singalsu: I wonder if there's a suitable code-block style for generic ASCII stuff. Do you have a recommendation? Just pasting the content without code-block would apply text formatting to it that would mess it up.

singalsu:
> these are great pictures in frequency domain, and very helpful for audio algorithm developers. Can we follow up with some pictures in time domain with sweep signal, it is more straightforward for non-audio algorithm developers, it will directly show which part of freq was impacted in a sweep signal.

That could be easily added to EQ tool, e.g. just show the resulting lin or log chirp waveform of the design. It would help to understand the dB -> linear amplitude impact.

algos/eq/equalizers_tuning.rst
mmaka1: All plantuml diagrams are already rendered to SVG and I have not seen any issues with web browsers reported.

algos/eq/equalizers_tuning.rst
mmaka1: You may replace it with (note the `::` at the end of the previous paragraph)
```
The settings file::

   The content of the file (each line indented)
   that is not processed and rendered using monospace font.
```

lgirdwood (approved):

PR #431, opened 2022-08-25
zephyr build guide: modify zephyr build dependency for better follow.

btian1:
Signed-off-by: Baofeng Tian <baofeng.tian@intel.com>

### DIFF
--- a/getting_started/build-guide/build-with-zephyr.rst
+++ b/getting_started/build-guide/build-with-zephyr.rst
@@ -17,10 +17,12 @@ This guide describes how to build and run |SOF| as a Zephyr application.
 Prepare
 *******
 
-- The easiest way to build Zephyr is to use its recommended toolchain which is included in its SDK. Refer to `Install Zephyr SDK <https://docs.zephyrproject.org/latest/getting_started/index.html#install-zephyr-sdk>`_ for details.
-
 - Install **west**. Zephyr uses west as a source management and building system. Follow
-  the Zephyr `Getting Started <https://docs.zephyrproject.org/latest/getting_started/index.html#>`_ guide for dependencies and for the west installation.
+  the Zephyr `Getting Started <https://docs.zephyrproject.org/latest/getting_started/index.html#>`_ guide for dependencies and for the west installation, for sof zephyr build, just need follow with below chapters:
+  Select and Update OS
+  Install dependencies
+  Get Zephyr and install Python dependencies
+  Install Zephyr SDK
 
 Clone and initialize SOF project
 ********************************


getting_started/build-guide/build-with-zephyr.rst
```diff
@@ -17,10 +17,12 @@ This guide describes how to build and run |SOF| as a Zephyr application.
...
 
-- The easiest way to build Zephyr is to use its recommended toolchain which is included in its SDK. Refer to `Install Zephyr SDK <https://docs.zephyrproject.org/latest/getting_started/index.html#install-zephyr-sdk>`_ for details.
-
 - Install **west**. Zephyr uses west as a source management and building system. Follow
-  the Zephyr `Getting Started <https://docs.zephyrproject.org/latest/getting_started/index.html#>`_ guide for dependencies and for the west installation.
+  the Zephyr `Getting Started <https://docs.zephyrproject.org/latest/getting_started/index.html#>`_ guide for dependencies and for the west installation, for sof zephyr build, just need follow with below chapters:
+  Select and Update OS
+  Install dependencies
+  Get Zephyr and install Python dependencies
+  Install Zephyr SDK
```
lgirdwood, line 25: @btian1 It sounds like the deleted Zephyr SDK links at the top are confusing for readers ? i.e. they need to follow the Zephyr getting started guide first. But do we need this extra teext below it ? @deb-intel any advice ?

deb-intel: We need to keep this as simple as possible but we don't want to confuse them. I don't see the need to tell them to install west at the very beginning of this page when it is already described as a step in Zephyr's Getting Started instructions. 

Unless Zephyr's instructions include something the user does not need to do, then just tell them to install Zephyr by following the instructions on the Getting Started page (provide the link). Mention that it includes dependencies that they must also install, particularly west.  

It seems to me that the really really important thing users need to know is that the west manifest might not initialize correctly because of the info you describe in "Clone and initialize SOF project." If you could mention this ahead of time, that would be useful although I don't know how they can avoid the conflict if installing Zephyr requires creating a west manifest which they have to do before they initialize SOF.

btian1: two main point is: 
1. Zephyr SDK guide is included in second getstarted link, you can check.
2. in get started link, there are much more than SOF need, so I suggest to install until Install Zephyr SDK.

lgirdwood: @kv2019i @marc-hb any comments here, you guys have been using west. I do now think it makes sense to install west as first step once we have the submodules removed (but it's not yet needed for xtos ?).

marc-hb (commented):
I don't understand what was wrong, the commit message is empty and "for better follow" does not explain any problem.

The current version does NOT ask to follow the entire Zephyr's Getting Started guide. It only points at two specific steps. This was just updated by @aborisovich and reviewed.

btian1:
@marc-hb , this doc is for sof, what I am want to share is:

Zephyr SDK install guide is included in  getstarted link, so removed the first line in this patch.

In getstarted, chapter 4:
  Install Zephyr SDK

Thanks
Tim

marc-hb:
> there is nothing wrong, but repeated for install Zephyr SDK, agree?

No, it's not just a "repeat". The Zephyr SDK is a "special" dependency because many people must not use the Zephyr SDK but some other, proprietary SDK depending on which particular product they're working on. That's why it deserves a special mention here.

PR #432, opened 2022-09-19
SOF-Zephyr Firmware Architecture chapters

mwasko:
SOF firmware build with Zephyr has major architecture and interface differences in compare to XTOS version which require documentation.

All the existing firmware chapters have been grouped in SOF-XTOS section and next to it the new SOF-Zephyr section was created.

This PR will be followed with more SOF-Zephyr documentation.


dbaluta:
@all I wonder if we can add a job that publishes the rendered documentation so that we can easily review it before it gets merged?

mwasko:
> @ALL I wonder if we can add a job that publishes the rendered documentation so that we can easily review it before it gets merged?



> @ALL I wonder if we can add a job that publishes the rendered documentation so that we can easily review it before it gets merged?

Publishing is probably not an option, but having a build job that expose output for download could be enough to open the generated html locally. However the restructured text format is pretty straight forward for review. The only problem are plantuml diagrams, but here you can use Visual Studio Code plantuml plugin to preview or try some online plantutml generators.

plbossart:
@marc-hb mentioned something a while ago to directly see the rendered result, of course I forgot but what you want @dbaluta already exists.

marc-hb:
The "Files Changed" tabs has very good rendering features hidden behind obscure/cryptic icons.

Take a closer look at the icons in the top-right corner of each file. The document icon has "richer diff": this is a diff of the rendered content. Hidden behind the triple-dots you have the usual "View File" feature EXCEPT Github detects the file is `.rst` and renders it.

To be honest I never remember which button does what. I only remember to go to the "Files Changed" tab and try them all until I find what I want. There's a small number of them.

The rendering is not exactly the same as the final, rendered HTML (for a start it renders files in isolation) but for reviewing small changes it's much more than good enough.

marc-hb:
> https://github.com/ALL I wonder if we can add a job that publishes the rendered documentation so that we can easily review it before it gets merged?

It should be possible for:
- some amount of work. Any volunteer?
- Probably some storage fee. Any.... volunteer? :-)

https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

I'm NOT volunteering because:
- I find Github's .rst rendering good enough for reviewing small changes, see above.
- For reviewing bigger changes I just fetch and compile it myself... which anyone should be able to do for a reasonably small effort. If not then we have a usability issue that should be fixed (I don't think we do, at least not on Linux).

architectures/firmware/index.rst
```diff
@@ -6,14 +6,7 @@ Firmware Architecture
...
 
-   overview
-   mem-mgmt
-   pm-runtime/index
-   schedulers
-   drivers/index
-   components/index
-   pipelines/index
-   kd_integration/index
+   sof-legacy/index
```
lgirdwood, line 9: Probably better to say sof-xtos instead of legacy as xtos still used. Fwiw, I intend to do the same name fix in FW for v2.4.

lgirdwood (commented):
Best if we use "xtos" instead of "legacy". We can then use legacy when no one is using xtos.

architectures/firmware/index.rst
mwasko: @lgirdwood legacy naming has been updated with xtos

deb-intel:
@lgirdwood @mwasko I'm looking at all page changes (a lot!) to see if any filename changes caused a break. Then I'll merge them and check again. Note that the Build and Deploy check did not pass. This might be similar to an earlier situation but perhaps it will merge anyway.

deb-intel (approved):
will merge and make updates in an a new PR.

PR #440, opened 2022-11-04
FW Architecture chapter that describe MPP scheduling using Zephyr infrastructure

mwasko:
The chapter describes MPP scheduling flows, task types and their usage in SOF based on Zephyr API.


lgirdwood (commented):
I think we should put the development state next to each scheduler type i.e. we know LL has been upstream since version 1.x, but DP is WIP and TwB are WIP. This way no one will read the docs and think they can use TwB today.

lgirdwood:
@marc-hb I think we are missing a sphinx package ?

mwasko:
> I think we should put the development state next to each scheduler type i.e. we know LL has been upstream since version 1.x, but DP is WIP and TwB are WIP. This way no one will read the docs and think they can use TwB today.

The approach with development state sounds good. I will update the PR with the proper information.

lgirdwood:
@

> @marc-hb I think we are missing a sphinx package ?

@deb-intel is there anyone who could comment on the sphinx config ?

deb-intel:
> 

@lgirdwood @marc-hb I'm not sure why this particular build is looking for sphinxcontrib.blockdiag (which is causing the failure) in this PR but not in more recent PRs. @intelkevinputnam, would you mind taking a quick look at this?

marc-hb:
I'll fix this today, sorry for the delay.

marc-hb:
Finally got to this sorry.

So I don't need to fix anything because it is exactly as @intelkevinputnam said.

@mwasko please add the new blockdiag requirement to `requirements-lax.txt` too and that should be it.

lgirdwood (approved):

PR #456, opened 2023-03-29
arch: fw: cleanup links in memory management overview

mmaka1:
Initial description was unbalanced. One section contained links while the other had the links moved to the Read More section below the diagram.

Now all the links are moved to the Read More. External links are annotated to let the reader know that opening results in move to another website.


mwasko (approved):

kv2019i (approved):
Thanks @mmaka1 !

lgirdwood (approved):

PR #466, opened 2023-05-03
sof-logger-doc: how to disable dsp suspension on imx8mp

eurofun:
plus some general restructuring

### DIFF
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -150,6 +150,9 @@ Examples
   file and start reading from position 0 and thus be in sync with the firmware
   when it is resumed.
 
+  Sometimes error messages about dropped logs are printed. If that is a problem,
+  increasing DMA_TRACE_LOCAL_SIZE in the relevant platform.h file can be helpful.
+ 
 Trace filtering
 ***************
 
--- a/developer_guides/debugability/logger/index.rst
+++ b/developer_guides/debugability/logger/index.rst
@@ -228,21 +228,23 @@ Disabling DSP power gating
 
 After a firmware reset (such as after power gating in suspend mode) custom filter settings will be lost.
 Thus consider disabling power gating during your debug session. The way this is done is slightly different on every platform,
-Consider disabling power gating during your debug session by entering the following:
+two examples follow:
 
-.. code:: bash
+1. Intel PCI based:
+
+.. code-block:: bash
 
   sudo su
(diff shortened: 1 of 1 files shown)


paulstelian97:
Hello. The changes themselves are good. However, there's a few structural things that prevent us from using your work as-is:

1. The commits aren't properly separated. Each commit should only do one thing. I still expect 3 commits, but one of them only does the info of disabling automatic suspend on imx8mp, one only adds (or rewrites) the information about trace verbosity flags, and one does the various smaller cleanups (such as shorter lines, fixing the underlines on titles etc).
2. The commit titles and descriptions are... lacking. I did see some potentially good stuff (e.g. the "logger:" prefix makes sense for all 3 commits). A commit should have a title and a description. The title is a summary of the changes, and the description is a more detailed explanation. For example, on the commit that does the cleanup you can put in the description the individual cleanups (underlines + line wrapping etc).
3. None of your commits has the COA. This is a problem because without that we cannot use your work at all. The COA implicitly states that the changes were authored by you, and that you allow us to use them. A proper COA looks something like:

```
Signed-off-by: Full Name <mail.address@mail.domain>
```

You can look at past commits for examples. If there are multiple authors which all had a significant contribution to the commits you can add multiple such lines, one for each of the authors.

developer_guides/debugability/logger/index.rst
```diff
@@ -204,19 +204,27 @@ Instance descriptions can have one of the following forms:
...
-   sof-loggerr -l ldc_file -t -Fc=* -Fv=*1.*
+   sof-logger -l ldc_file -t -Fc=* -Fv=*1.*
 
-.. note::
-  To track a verbose message, select the "Trace verbose" option under the "Trace" menu from the firmware build.
+Verbose and debug log levels
+----------------------------
+
+To enable verbose and debug trace messages, select the "Trace->Trace verbose" option in the firmware build
+menuconfig (in addition to setting the proper log levels as described above).
```
paulstelian97, line 224: This still needs separate commits. One commit that does the rephrasing/separation of info, and one that fixes whitespace/typos/etc. Also the commit message being in all lowercase isn't ideal.

paulstelian97 (commented):
Already much better, still needs work.

eurofun:
Thanks for the critique, I updated the commits accordingly. I even have a typo in the branch name, but I'll leave that as-is...

dbaluta (approved):

dbaluta:
@eurofun this looks good to me! Thanks for the patches.

@deb-intel @lgirdwood  Good for you?

deb-intel (approved):
LGTM. Nice job.

PR #481, opened 2023-12-21
getting_started: intel_debug: Update the sof-dyndbg.conf example

ujfalusi:
Add snd_sof_intel_hda_generic and remove snd_sof_intel_ipc to the list

### DIFF
--- a/getting_started/intel_debug/suggestions.rst
+++ b/getting_started/intel_debug/suggestions.rst
@@ -135,7 +135,6 @@ file:
    options snd_intel_sdw_acpi dyndbg=+pmf
 
    # SOF internals
-   options snd_sof_intel_ipc dyndbg=+pmf
    options snd_sof_intel_hda_common dyndbg=+pmf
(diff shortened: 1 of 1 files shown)


getting_started/intel_debug/suggestions.rst
```diff
@@ -135,8 +135,8 @@ file:
    options snd_intel_sdw_acpi dyndbg=+pmf
 
    # SOF internals
-   options snd_sof_intel_ipc dyndbg=+pmf
    options snd_sof_intel_hda_common dyndbg=+pmf
+   options snd_sof_intel_hda_generic dyndbg=+pmf
```
lgirdwood, line 139: Should we say from whence ? i.e. from 6.6+ do this else do that.

ujfalusi: I will separate the two change.
I cannot find any indication that we ever had `snd_sof_intel_ipc`

plbossart (approved):

lgirdwood (approved):

Issue #472, opened 2023-07-10
'ImageDraw' object has no attribute 'textsize' warning with new pillow version 10.0.0


marc-hb:
On July 1st, python library "Pillow" (needed by sphinxcontrib.blockdiag) released version 10.0.0
https://pypi.org/project/Pillow/#history

The sof-docs build immediately started to fail with the warning below.

https://github.com/thesofproject/sof/actions/workflows/daily-tests.yml
https://github.com/thesofproject/sof/actions/runs/5433906267/jobs/9881925923

> WARNING: dot code 'blockdiag edf_scheduling {\n\n\tnode_width = 250;\n\tnode_height = 120;\n\tdefault_fontsize = 16;\n\n\tComp_1 -> Comp_2\n\tcomment_1 -> Comp_2 [style=dashed]\n\tComp_2 -> Comp_3\n\tcomment_2 -> Comp_3 [style=dashed]\n\tComp_3 -> Comp_4\n\tcomment_3 -> Comp_4 [style=dashed]\n\tComp_4 -> sink\n\tcomment_4 -> sink [style=dashed]\n\n\tComp_1 [label="DP component 1\\n\n\t\t*processing period\\n\n\t\t*compute requirement"]\n\tComp_2 [label="DP component 2\\n\n\t\t*processing period\\n\n\t\t*compute requirement"]\n\tComp_3 [label="DP component 3\\n\n\t\t*processing period\\n\n\t\t*compute requirement"]\n\tComp_4 [label="DP component 4\\n\n\t\t*processing period\\n\n\t\t*compute requirement"]\n\n\tsink [label="real time sink", shape=endpoint, fontsize = 16]\n\n\tcomment_1 [label="DP1 to deliver data let\\n\n\t\tDP2 meet its objective"]\n\tcomment_2 [label="DP2 to deliver data let\\n\n\t\tDP3 meet its objective"]\n\tcomment_3 [label="DP3 to deliver data let\\n\n\t\tDP4 meet its objective"]\n\tcomment_4 [label="DP4 to deliver data\\n\n\t\tto real time-sink"]\n}\n': 'ImageDraw' object has no attribute 'textsize'
generating indices... genindex done

Also seen in recent sof-docs PRs:
https://github.com/thesofproject/sof-docs/actions/runs/5461579335/jobs/9939764550

To reproduce:
```
pip3 install -r scripts/requirements.txt
# override the debian package python3-pil or whatever else
pip3 install pillow==10.0.0

[...]

PR #485, opened 2024-02-16
requirements.txt updates

deb-intel:
sphinxcontrib-applehelp was recently updated to 1.0.6. This caused a series of compatibility issues between sphinxcontrib-applehelp, sphinx, and breathe, which caused all recent builds to fail. To lessen this problem both now and in the future, we removed hardcoded software versions in the requirements.tx file although we updated sphinx to 7.0 or higher. This will enable the system to pick compatible software versions to match our needs.


mwasko:
> sphinxcontrib-applehelp was recently updated to 1.0.6. This caused a series of compatibility issues between sphinxcontrib-applehelp, sphinx, and breathe, which caused all recent builds to fail.

It looks like a gap in the requirements list to point to specific sphinxcontrib-applehelp version?

> To lessen this problem both now and in the future, we removed hardcoded software versions in the requirements.tx file although we updated sphinx to 7.0 or higher. This will enable the system to pick compatible software versions to match our needs.

How this approach is going to guarantee that our SOF documentation will compile successfully with latest tools version?

lgirdwood:
> > sphinxcontrib-applehelp was recently updated to 1.0.6. This caused a series of compatibility issues between sphinxcontrib-applehelp, sphinx, and breathe, which caused all recent builds to fail.
> 
> It looks like a gap in the requirements list to point to specific sphinxcontrib-applehelp version?
> 
> > To lessen this problem both now and in the future, we removed hardcoded software versions in the requirements.tx file although we updated sphinx to 7.0 or higher. This will enable the system to pick compatible software versions to match our needs.
> 
> How this approach is going to guarantee that our SOF documentation will compile successfully with latest tools version?

IIRC, we align with the same versions used by the Zephyr docs (as SOF docs and Zephyr docs use the same framework). @deb-intel is my understanding still correct ?

deb-intel:
@mwasko @lgirdwood When @intelkevinputnam and I built SOF locally, the build set the proper requirements based on our setting Sphinx to >=7. It automatically new to use sphinxcontrib-applehelp version 1.0.4 instead of 1.0.6 as well as the best version of breathe. The build kept failing because we kept giving it specific versions. I think this will continue to happen every time breathe, docutils, and software we didn't even know about (such as -applehelp) makes updates.

deb-intel:
IIRC, we align with the same versions used by the Zephyr docs (as SOF docs and Zephyr docs use the same framework). @deb-intel is my understanding still correct ?

@lgirdwood I am not familiar with how Zephyr handles things. I know they no longer have a full-time doc person. I'll ask one of our guys if he can see what's up with their requirements.

intelkevinputnam:
I recommend moving to version 7 of sphinx. A good option for a specific version is 7.1.2 as this will have the broadest compatibility for those who might want to build it locally (later versions require Python 10 or better).

lgirdwood:
Thanks @deb-intel @intelkevinputnam for explaining.

lgirdwood:
@mwasko good for you now ?

mwasko (approved):

marc-hb:
> When @intelkevinputnam and I built SOF locally,

What is "locally"? Ubuntu, Fedora, other? Which version? Each distribution and even each user has a different combination. Versions were hardcoded in `requirements.txt` to solve that problem (We already have `requirements-lax.txt` for more flexibility)

> The build kept failing because we kept giving it specific versions.

In that case you just un-hardcode locally  and temporarily, find a complete set of versions that works (as you already did) and then snapshot and hardcode that combination. Problem solved for everyone and for a long time as opposed to just on one system for a short time. That has been working and it can work again.

> It looks like a gap in the requirements list to point to specific sphinxcontrib-applehelp version?

Yes of course. The solution was to hardcode MORE, not less.

> I think this will continue to happen every time breathe, docutils, and software we didn't even know about (such as -applehelp) makes updates.

It will for sure happen again very soon now that all the versions are un-harcode. In fact it's already happening:

- Using `requirements-lax.txt` I just hit this recent regression on my system: https://github.com/sphinx-doc/sphinx/issues/9433 People have been avoiding it by... hardcoding.

- `blockdiag` is now orphaned (#472). We have been (and still are!) avoiding that failure by hardcoding pillow (73ac8f55622d)

- etc.

PS: I spent days finding combinations that work, please copy me next time.

scripts/requirements.txt
```diff
@@ -1,11 +1,13 @@
 # This file hardcodes validated versions with '==',
 # see requirements-lax.txt for an alternative.
-breathe==4.29.2
-sphinx==4.5.0
-docutils==0.17.1
+
```
marc-hb, line 3: The comment still says "This file hardcodes..."

PR #493, opened 2024-04-11
developer-guide: add LLEXT module documentation

lyakh:
Describe LLEXT loadable module support in SOF.


developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+Adapter drivers, that would be built exclusively as loadable modules.
+
+Note, that a single LLEXT object file can also contain multiple Module Adapter
+modules. In that case an array of ``struct sof_module_api_build_info`` objects
+is needed and the TOML file should contain those multiple module entries too.
+
+Installation
+************
+
+As described in :ref:`intel_debug_introduction` the |SOF| Linux kernel driver
```
marc-hb, line 75: Unless I missed something that page does not really "describe" anything; it merely gives a /lib/firmware location.

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+loads SOF modules, using their UUIDs. For SOF internal modules the process
```
marc-hb, line 76: What does "using their UUIDs" mean? How does the kernel know which modules to load and which not?

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+of creation and installation of modules in a deployment tree is automated by the
+xtensa-build-zephyr.py script.
```
marc-hb, line 78: Is xtensa-build-zephyr.py optional? If yes, is there a way to build _only_ a module or a plan to support that? If not, how do you build with a different key?

If it is required to build everything at the same time with the same key then what's the purpose of LLEXT?

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+for detailed documentation. In short, the Zephyr Linkable Loadable Extensions
+(LLEXT) API implements support for run-time loading and unloading of ELF-format
+executable code and data.
+
+SOF Use of the LLEXT API
+************************
+
+SOF has multiple ways to implement loadable modules. LLEXT is one of them.
+With it modules are built as shared or relocatable ELF objects with an addition
+of a cryptographic signature and a manifest. When loaded and instantiated,
```
marc-hb, line 21: Signature with what key(s)? Same as base firmware, different, both?

developer_guides/firmware/llext_modules.rst
lyakh: @marc-hb Topology specifies UUIDs, the kernel driver also parses the base firmware manifest to see which UUIDs are built-in. For any, that are specified in topology but aren't available in the base firmware, an attempt will be made to load them dynamically. But this is generic SOF architecture information, not LLEXT specific

developer_guides/firmware/llext_modules.rst
lyakh: @marc-hb the purpose is not to have to load everything into SRAM at once. Only currently active pipelines need all modules in SRAM, the rest can be kept out one way or another.
xtensa-build-zephyr.py is optional in the sense, that it is "only" calling other programs to actually manipulate files, it isn't doing any deep binary manipulation itself. As you have seen in https://github.com/thesofproject/sof/pull/9013/commits/f85900353ce226a8093971f8bc91babd5debf2ba it "just" locates all the modules, built by cmake, calls rimage to sign them and installs the result into a deployment tree, including soft links. Yes, you can do that all manually, but it isn't very convenient or simple.
As for building only a module - you need a full Zephyr+SOF build environment - you need .config. It should be possible to sign modules with alternative keys, just call rimage manually as mentioned above.

developer_guides/firmware/llext_modules.rst
lyakh: @marc-hb wording updated

lgirdwood (approved):

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+A TOML configuration file is needed for building of LLEXT modules too. It is
+generated by the C preprocessor at build time from the same components, as would
+be used for a monolithic build. A small additional file to bring components
+together is needed too, similar to
+src/samples/audio/smart_amp_test_llext/llext.toml.h.
+
+Finally an additional CMakeLists.txt is needed similar to
+src/samples/audio/smart_amp_test_llext/CMakeLists.txt. Currently it contains a
+lot of open-coded instructions, later it will be ported to use Zephyr's
+``add_llext_target()`` function.
```
marc-hb, line 60: "In the future" in code reviews often means "never"  but it makes it all the way to the official documentation then it becomes a certainty :-D

Is `add_llext_target()` still not ready? Either way I think this is basically an implementation detail not really worth mentioning here - especially not if you intend to clean it up soon. A TODO in the code itself seems enough.

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+
+With that in place, it is also possible to switch between monolithic and modular
+builds by specifying the module as "tristate" in its Kconfig and selecting "m"
+for modular builds, however, for modules, exclusively built as LLEXT objects,
+this might not be necessary. Thus it is possible to implement third party Module
```
marc-hb, line 65: By "not necessary" you mean "not possible"? As in "exclusively". I think I'm guessing what you mean but this is confusing.

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,78 @@
...
+Adapter drivers, that would be built exclusively as loadable modules.
+
+Note, that a single LLEXT object file can also contain multiple Module Adapter
+modules. In that case an array of ``struct sof_module_api_build_info`` objects
```
marc-hb, line 69: Is EXPORT_SYMBOL() required between these? Or is that not supported yet?

This "multiple modules inside one module" does not look like a "note", more like a key design aspect. It should be moved near the top of this section.

developer_guides/firmware/llext_modules.rst
marc-hb: Thanks, I think I found it at `_apps-component-overview`, please add a link to that.

Earlier you mentioned that a single LLEXT file can have multiple modules, in that case which UUID is used for its filename?

developer_guides/firmware/llext_modules.rst
marc-hb: > It should be possible to sign modules with alternative keys, just call rimage manually as mentioned above.
> ...
> Yes, you can do that all manually, but it isn't very convenient or simple.

Please mention at least that so readers have at least some vague idea of what is "automated" and some high level idea of what problem xtensa-build-zephyr.py solves.

developer_guides/firmware/llext_modules.rst
lyakh: > Earlier you mentioned that a single LLEXT file can have multiple modules, in that case which UUID is used for its filename?

@marc-hb none. Filenames are created from a common name: "eq_iir," "mixin_mixout," etc. UUIDs are soft links to those files

marc-hb: Did you mean "all of them"? As in:
```
  UUID1.bin -> module1and2combined.llext
  UUID2.bin -> module1and2combined.llext
```

developer_guides/firmware/llext_modules.rst
lyakh: @marc-hb no, I didn't mean that. I don't know if anybody would want to write modules that they never need to build into the actual SOF image.

developer_guides/firmware/llext_modules.rst
lyakh: > Is EXPORT_SYMBOL() required between these? Or is that not supported yet?

no, that isn't needed. It's a single ELF image.

lgirdwood (approved):

deb-intel:
@marc-hb Are you satisfied with the updates?

developer_guides/firmware/llext_modules.rst
marc-hb: I'm still confused sorry. I guess I don't understand what is a "module exclusively built as LLEXT object". Sounds like you select "monolithic" but that is ignored by some modules? Why?

lyakh: @marc-hb no, that isn't what I meant. I just meant, that if you're sure to only build your code as a loadable LLEXT module, it doesn't matter what you do in your Kconfig, you can just hard fix it to always build as a module.

marc-hb: I still have no idea how you "hard fix" the build-as-module and how the Kconfig gets ignored in that case (which is confusing?)

lyakh: @marc-hb this is what I currently use to switch to building a loadable module: https://github.com/thesofproject/sof/blob/fdb67ee782808f6d5488a53441cdecd75f458677/zephyr/CMakeLists.txt#L874 If you only ever build your code as such, you don't need that `if`, you use that code unconditionally.

lgirdwood:
@marc-hb ping

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,89 @@
...
+containing a ``struct sof_man_module_manifest`` object. The latter should also
+contain a pointer to a module entry point function, returning a pointer to the
+module's ``struct module_interface`` instance. All these additions can be
+performed, using ``SOF_LLEXT_MOD_ENTRY()``, ``SOF_LLEXT_MODULE_MANIFEST()`` and
+``SOF_LLEXT_BUILDINFO`` helper macros. See src/audio/eq_iir/eq_iir.c for an
+example.
+
+A TOML configuration file is needed for building of LLEXT modules too. It is
+generated by the C preprocessor at build time from the same components, as would
+be used for a monolithic build. For this preprocessor run a small header file,
```
marc-hb, line 62: I'm not a "English comma expert" but I really struggled to parse this sentence (also because `run` is both a noun and a verb). I think this is what it means:
```suggestion
be used for a monolithic build. For this preprocessor run, a small header file that ...
```

developer_guides/firmware/llext_modules.rst
```diff
@@ -0,0 +1,89 @@
...
+**************************
+
+At the moment only modules, implementing the Module Adapter API
+:ref:`apps-comp-world` are supported.
+
+It is possible to implement multiple Module Adapter modules with a common code
+base, i.e. sharing a set of source files and functions. Then a single LLEXT
+object would be created, implementing multiple Module Adapter interfaces. In
+that case an array of ``struct sof_module_api_build_info`` objects is needed and
+the TOML file should contain those multiple module entries too.
```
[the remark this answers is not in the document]
lyakh: @lgirdwood you mean when exporting symbols between LLEXTs? Yes, now that https://github.com/zephyrproject-rtos/zephyr/pull/77473/commits/d59650906e98031a4e38d5a72fac82a671122eeb has been merged Zephyr only should use `EXPORT_SYMBOL()` for both cases - exporting from main firmware and from modules.

lgirdwood (approved):

PR #504, opened 2024-11-21
update admin and tsc members

cujomalainey:
It has been great working on the project, I will miss it. Maybe we will cross paths in the zephyr tree.


lgirdwood (approved):
@johnylin76 we need to add 1 other representative from Google.

PR #506, opened 2024-12-16
intel_debug: introduction: Split firmware lookup table into two for m…

ujfalusi:
…onolithic and modular

Separate the firmware lookup table for monolithic and modular SOF release to be able to document the locations and file names the firmware will be looking for the individual files.

Extend the description of the two type of SOF release and convert the list-table to a normal table for better descriptions for the configurations.

### DIFF
--- a/getting_started/intel_debug/introduction.rst
+++ b/getting_started/intel_debug/introduction.rst
@@ -127,7 +127,7 @@ configuration issue.
 
 An IPC4 library is a container of a single or multiple modules (bundle) which
 can be loaded to the firmware after it is booted up.
-Library loading is  supported on Meteor Lake (ACE1) or newer platforms.
+Library loading is supported on Meteor Lake (ACE1) or newer platforms.
 
 Background information: the base firmware always resides in DSP SRAM while the
 loaded library is stored in DRAM memory and only the needed code is copied to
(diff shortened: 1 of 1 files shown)


getting_started/intel_debug/introduction.rst
```diff
@@ -142,16 +142,24 @@ See :ref:`llext_modules` for technical details.
...
 
-1.3.1. Modular SOF releases
----------------------------
+1.3.1. Monolithic SOF releases
+------------------------------
 
-See :ref:`loadable-libraries` for details about library support in general.
+The release contains:
+ - **sof-PLAT.ri** : The firmware binary
+ - **UUID.bin** : Mainly 3rd party libraries identified by UUID. If the library contains multiple modules then a UUID symlink must be provided for each one. Optional, only supported with IPC4 and on Meteor Lake and newer platforms.
```
lyakh, line 150: should this really still be called "monolithic?" To me "monolithic" is when you have no loadable modules at all, no run-time code loading. Do we need to document 3 types then? Monolithic, individual modules and bundled modules?

ujfalusi: I think it was a mistake to release the DRC as module in a tech demo style for LNL, but my point here is:
monolithic: basefw (+optional UUID libraries, if any)
modular: basefw + openmodules + debug (+optional UUID libraries, if any)

The item tells that the UUID.bin is only supported on MTL+, which is correct.

getting_started/intel_debug/introduction.rst
```diff
@@ -160,41 +168,46 @@ Notes:
...
-     - IPC4
-     - /lib/firmware/intel/sof-ipc4-lib/PLAT/
-     - PLAT = mtl, lnl, ...
-   * - Meteor Lake and newer Loadable libraries (community signed)
-     - IPC4
-     - /lib/firmware/intel/sof-ipc4-lib/PLAT/community/
-     - PLAT = mtl, lnl, ...
+Linux SOF will look up firmware files at the following paths.
+
+Look-up paths per Intel platform for **mopnolithic releases**
```
kv2019i, line 173: Not sure world is ready for **mopnolithic** releases yet...?

getting_started/intel_debug/introduction.rst
```diff
@@ -160,41 +168,46 @@ Notes:
...
++-----------------------------------------------------------+--------+------------------------------------------------+-----------+-----------------------------------+
+
+Look-up paths per Intel platform for **modular releases (IPC4 only)**
+
+.. _intel_modular_firmware_paths:
+
++-----------------------------------------------------------+------------------------------------------------+-------------------------+--------------------+
+|Platform                                                   |Load path                                       |File name                |Notes               |
++===========================================================+================================================+=========================+====================+
+|Meteor Lake and newer                                      |/lib/firmware/intel/sof-ipc4/PLAT/              ||                        |PLAT = mtl, lnl, ...|
```
kv2019i, line 200: We already know we won't support this for MTL, LNL, so why mention these at all?

kv2019i (commented):
Naming is hard. Some comments inline...

getting_started/intel_debug/introduction.rst
ujfalusi: Don't say never... Technically it is possible.

ujfalusi:
Changes since v1:
- fix typo mopnolithic -> monolithic
- Clarify that the modular release is PTL+ targeted (Moving MTL/LNL to this mode can break users system with older kernels)

getting_started/intel_debug/introduction.rst
kv2019i: @ujfalusi  Github ate my comment (or user error), but I did intend to comment at this as well. I get @lyakh 's concern here, were are in fact shipping a modular build in 2024.09 , so it is what it is. 

I do like the "Modular Release" name, so not sure how to name the "before" and "after" layouts. Maybe just call it what is, "SOF Modular Release Layout V2 (for PTL and newer)", and "SOF Modular Release Layout V1" (for LNL/ARL/MTL)". Not pretty but describes what it is.

ujfalusi: I consider the UUID based loading as monolithic+library. The LNL is a short sighted tech demo to show that we can load a single module (as library) to the firmware.

But technically MTL+ will be able to load modular release (basefw+openmodules+debug+UUID), it is just that sof-bin will only going top ship modular release for PTL+.

My distinction criteria is:
If there is basefw _and_ openmodules _and_/_or_ debug) in the firmware path then it is a modular release, if there is only the basefw in there, it is monolithic.

The kernel will not care about any of these platform names, it will try to see if these fixed named libraries are there and if they are, load them.

and the UUID based library loading is also supported from MTL+ So, on PTL also.

Monolithic release == single library in fw path
Modular release == multiple libraries in fw path

Note: basefw is a library.

getting_started/intel_debug/introduction.rst
```diff
@@ -160,41 +168,46 @@ Notes:
...
+|                                                            |                                                || sof-PLAT-openmodules.ri|               |
+|                                                            |                                                || sof-PLAT-debug.ri      |               |
++------------------------------------------------------------+------------------------------------------------+                         |            -- |
+|Panther Lake and newer (community signed)                   |/lib/firmware/intel/sof-ipc4/PLAT/community/    |                         |               |
++------------------------------------------------------------+------------------------------------------------+-------------------------+            -- |
+|Panther Lake and newer Loadable libraries                   |/lib/firmware/intel/sof-ipc4-lib/PLAT/          |UUID.bin                 |               |
++------------------------------------------------------------+------------------------------------------------+                         |            -- |
```
kv2019i, line 208: @ujfalusi Also, why not just put UUID.bins to the main directory? The UUID namespace is unique by definition, so what do we gain by putting these to sof-ipc4-lib versus sof-ipc4?

ujfalusi: I cannot find the thread on why we ended up with a separate directory for the libraries, it is mentioned here at least: https://github.com/thesofproject/linux/pull/3826#discussion_r949059248

I think it has something to do with the fact that libraries can be signed with a vendor key and in that case they cannot be alongside of the Intel or community signed basefw, but the basefw can be only signed with Intel or community key.

But this has been the rule ever since we have the UUID based library lookup. Initially I wanted to toss them alongside of the basefw, but then we have changed direction and there were a good reason to do that (but what? I cannot recall).

kv2019i: Ack, thanks @ujfalusi . So this goes to "water under the bridge" category, let's stick to what we've done already and be consistent with previous IPC4 FW releases.

kv2019i (approved):
Some grumblings on "monolithic release" term, but I don't have better alternatives, so +1 for this.

getting_started/intel_debug/introduction.rst
ranj063: I agree with the name monolithic being a bit of a misnomer too. It hints at not supporting loadable modules.

I have a suggestion, how about legacy(for the old releases) and bespoke (for the multiple binaries version) releases?

ujfalusi:
Changes since v2:
- rename the monolithic to single firmware.

ranj063:
> What about this naming: 1.3.1 Single firmware release + UUID on demand library support on MTL+ 1.3.2 Split firmware release (basefw+openmodules+debug) + UUID on demand library support
> 
> instead of the Single/Modular.

or how about unified/fragmented? I'm good with single/split too.

ujfalusi:
Changes since v3:
new terms to use: Non-modular and Modular SOF:
non-modular: No loadable library support (IPC3, IPC4 prior to MTL)
modular: supports library loading, two distribution mode: single firmware and split firmware

lyakh:
> What about this naming: 1.3.1 Single firmware release + UUID on demand library support on MTL+ 1.3.2 Split firmware release (basefw+openmodules+debug) + UUID on demand library support
> 
> instead of the Single/Modular.

@ujfalusi I like it! I was about to propose something like multi-part / single-part / modular, but split is ok too!

kv2019i (approved):
Even better!

ranj063 (approved):

lgirdwood (approved):

PR #511, opened 2025-03-06
zephyr-build-guide: define ZEPHYR_WORKSPACE before use

SurajSonawane2415:
`ZEPHYR_WORKSPACE` was not defined before use. This update ensures it is defined first, similar to how `SOF_WORKSPACE` is defined before use in **"Build toolchains and SOF from sources"**.

#### Testing  
Generated the HTML documentation locally using:  
       `make -C sof-docs VERBOSE=1 html`

Screenshot from 2025-03-06 15-51-35

I would appreciate your review and any suggestions for improvement. Thank you!


getting_started/build-guide/build-with-zephyr.rst
```diff
@@ -31,6 +31,7 @@ Initialize the west manifest ``$ZEPHYR_WORKSPACE/sof/west.yml`` using the ``west
 
       .. code-block:: bash
 
+         ZEPHYR_WORKSPACE=~/zephyr_workspace
```
lgirdwood, line 34: I think this is usually defined as ```~/work/sof/zephyr```

SurajSonawane2415: @lgirdwood Yes you are correct! I looked into other build instructions, including **build-from-scratch** and **build-with-docker** to align the Zephyr path with this build guide—the correct path should be: ZEPHYR_WORKSPACE=~/work/sof/zephyr

lgirdwood (approved):

PR #512, opened 2025-03-08
python: requirements: update Pillow, Sphinx

lgirdwood:
Pillow 11.1 seems to work with Sphinx 8.2. No build errors. Update.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -32,7 +32,7 @@ jobs:
   supported-reqs:
 
     name: 'Supported scripts/requirements.txt'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


lgirdwood:
Testing - if this works here I'll update commit message. Tested on Ubuntu 24.10

singalsu (dismissed):

jsarha (commented):
So this requirement is only for non pip environments? In general I'd hope that latest LTS (24.04.x) would be enough to do SOF development.

lgirdwood:
FW side PR is https://github.com/thesofproject/sof/pull/9982

lgirdwood:
> So this requirement is only for non pip environments? In general I'd hope that latest LTS (24.04.x) would be enough to do SOF development.

Yes - but recommendation is to use a python venv and use pip here.

kv2019i (dismissed):

lgirdwood:
Pushed to restart CI

lgirdwood:
Closing this PR as requested.

